PR #14, opened 2012-09-10
Add locking functionality

grooverdan:
This code adds the option locking to filesystem so  that consistent content of calendars is given. It uses an external package locking. If you're ok with this I'll happily update the website documentation.

Cheers

Daniel.
PS: also noticed git_folder in radicale/config.py ? Is this something in progress. I'm definitely interested in a git based storage and may implement this regardless.

### DIFF
--- a/config
+++ b/config
@@ -88,6 +88,10 @@ type = None
 # Storage backend
 type = filesystem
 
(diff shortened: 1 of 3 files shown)


liZe:
Locking has been implemented in #402.

PR #118, opened 2014-02-25
added bcrypt password hashing method

luqasz:
This pull request adds bcrypt password hashing method.
bcrypt library can be found here
https://code.google.com/p/py-bcrypt/
or installed via pip


liZe:
Fixed by #297.

PR #184, opened 2014-07-15
Radicale on GoogleAppEngine

erpic:
This code seems to work (no heavy testing).

I have used #FIXME: tags when I had questions.

I'd be happy to review this with someone. We could probably do that pretty quickly over the phone?
---
Update 16/7:

I have restructured my code to allow for better insertion of application-specific logic through subclassing ItemContainerAppengine and CollectionContainerAppengine and also added pre_* and post_* hooks for operations on items.

As discussed on the mailing list, this pull request is for general comments on the proposed architecture changes, not to be analyzed line by line for direct commit into the main branch. 

The most efficient way would be a quick discussion on the phone. I am based in France. Thanks.


liZe:
I think that we can close this pull request now and keep this ticket for the record.

PR #187, opened 2014-07-20
Implementing group restrictions

bouttier:
This patch add an new optional option 'group' in rights configuration file.
If specified, group membership of the user is verified before access is granted.
This need an auth backend implementing group membership. If not, access is
denied (security first). This patch implements it in PAM backend ;
implementation in LDAP backend is planned.

### DIFF
--- a/radicale/auth/PAM.py
+++ b/radicale/auth/PAM.py
@@ -91,3 +91,44 @@ def is_authenticated(user, password):
             GROUP_MEMBERSHIP)
 
     return False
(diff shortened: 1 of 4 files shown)


Unrud:
All authentication modules besides `htpasswd` were removed in Radicale 2.0.0.
Other methods can be implemented as [plugins](http://radicale.org/plugins/).
The group restrictions can be implemented in combination with a rights plugin.

PR #199, opened 2014-08-09
Fixed propfind handling

muggenhor:
This expands PROPFIND support to fully all three variants of PROPFIND as mentioned in [RFC 4918 section 9.1](https://tools.ietf.org/html/rfc4918#section-9.1).

Specifically it adds handling of `<propall/>` and `<propnames/>` requests.


liZe:
Thank you!

PR #200, opened 2014-08-09
Add Dovecot auth

muggenhor:
This adds code to authenticate against a Dovecot server *without* resorting to IMAP. It uses [Dovecot's Authentication Protocol][1] to communicate directly with Dovecot through a local/unix socket. This is functionally similar to the Courier auth module.

The primary advantage of using this instead of the IMAP module is that a CalDAV sync doesn't cause syslog to be spammed with lots of login/logout events in a short amount of time (I've seen bursts of 10/sec sustained for about a minute during initial download of a calendar).

[1]: http://wiki2.dovecot.org/Design/AuthProtocol

### DIFF
--- a/config
+++ b/config
@@ -73,7 +73,7 @@
 [auth]
 
 # Authentication method
-# Value: None | htpasswd | IMAP | LDAP | PAM | courier | http | remote_user | custom
+# Value: None | htpasswd | IMAP | LDAP | PAM | courier | dovecot | http | remote_user | custom
 #type = None
 
 # Custom authentication handler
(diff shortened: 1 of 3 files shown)


mrueg:
:+1: it would be great to see this included.

jbergstroem:
Could you make this python3 compatible as well?

jbergstroem:
(re python 3: buffers/strings)

muggenhor:
@jbergstroem I don't have the time right now nor do I see myself having it in the foreseeable future. Feel free to take my changes and make it Python 3 compatible yourself, or alternatively close this pull request.

liZe:
As it's defined in #372, the authentication modules have been removed from Radicale and will be available thanks to external "plugins". Is anyone interested in developing the plugin and/or maintaining it? You can add your comments in #391.

PR #201, opened 2014-08-09
MKCOL: don't crash when trying to create a "regular" collection

muggenhor:
Currently when trying MKCOL through a regular WebDAV client such as cadaver I get an exception and a traceback instead of a decent client error (4xx). This fixes that by considering a missing request-body for MKCOL to be HTTP/WebDAV precondition failure and reports it to the client as such.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -481,6 +481,9 @@ class Application(object):
 
         collection = write_collections[0]
 
+        if content is None:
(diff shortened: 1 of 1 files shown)


untitaker:
Why not handle MKCOL correctly? Given that GET requests already autocreate collections, this shouldn't be too hard.

liZe:
It's now correctly handled, thanks!

PR #202, opened 2014-08-09
IMAP: don't spam the logs about non-SSL connections to localhost

muggenhor:
When using IMAP auth with localhost as an IMAP server it is *not* unreasonable *not* to use SSL. (i.e. it is reasonable not to use SSL/encryption). Unfortunately the current logging strategy doesn't reflect that because on average 50% of the log messages coming from Radicale (with level=INFO set in log config) are about "transmitting unencrypted credentials".

This change reduces the spaminess of that. Although I personally don't think it would be a bad idea to remove the message altogether in case of localhost, there might be some servers configured with 'localhost' referring an external host where this message might be relevant. Thus my reason for leaving the message in while only reducing its frequency.


liZe:
Thanks!

PR #236, opened 2014-11-30
Catch OSError instead of IOError

untitaker:
If the collection doesn't exist yet, OSError(2, 'No such file or
directory') is raised.

https://travis-ci.org/untitaker/vdirsyncer/jobs/42540595

### DIFF
--- a/radicale/storage/multifilesystem.py
+++ b/radicale/storage/multifilesystem.py
@@ -29,6 +29,7 @@ import sys
 
 from . import filesystem
 from .. import ical
+from .. import log
 
 
 class Collection(filesystem.Collection):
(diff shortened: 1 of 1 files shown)


untitaker:
@liZe Why can't this be merged before 1.0?

liZe:
Here's what I get for `import codecs;codecs.open('test').read()` (from `filesystem.py`) with no `test` file:
- Python 2.7: `IOError`
- Python 3.4: `FileNotFoundError` (inherits from `OSError`)

We can probably get different errors with `filesystem` and `multifilesystem`, so what do you think of catching both?

untitaker:
Hmm, I think the `IOError` should be catched inside the loop.

untitaker:
Hang on, I'll update my PR.

untitaker:
@liZe I think this is good as-is.

Also I get test failures (from my own testsuite), I'll look into it.

PR #249, opened 2015-01-04
Change distutils to setuptools

jspricke:


liZe:
Is there a reason for that? I'm a bit lost in the distutils/setuptools war, and I don't see the advantages of setuptools for Radicale.

jspricke:
With setuptools you can do `python setup.py develop --user` which only adds links to the git repo, so you can test your development without having to call install every time.

untitaker:
``pip`` depends on ``setuptools``. I think it's safe to argue on that basis that it's now the officially endorsed package to import the ``setup`` function from.

liZe:
I'm convinced!

PR #255, opened 2015-01-23
Fix FutureWarning, explicitly test prop_element for None

jspricke:


untitaker:
@jspricke Could you post the warning you got?

jspricke:
Not sure how I triggered that back then, but it should be this one:
http://stackoverflow.com/questions/18583162/difference-between-if-obj-and-if-obj-is-not-none.

jelmer:
/home/jelmer/Radicale/radicale/xmlutils.py:469: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  props = [prop.tag for prop in prop_element] if prop_element else []

PR #260, opened 2015-02-09
change GET response from GONE to NOT FOUND when item is not found

deronnax:
I think "GONE" is really a bad return code. It's really little used in real world because it implies keeping track of deleted items which 99% of server don't do, and its semantic is the same as 404 and means no real difference to clients.
As Radicale doesn't keep track of deleted items and only knows that they are not found, I strongly suggest to replace GONE with NOT FOUND :)


untitaker:
:+1:

PR #280, opened 2015-04-20
Add compatibility patch for .authenticate().

fatlotus:
Current version of python-pam (https://pypi.python.org/pypi/python-pam/1.8.1) have changed the API slightly; this patch fixes these bugs.


liZe:
I didn't know that the pam module had been forked and now works with Pyhon 3, thanks!

PR #281, opened 2015-04-25
Fix compatibility for PAM auth

untitaker:
PR #280 invoked `pam.authenticate().authenticate()` for older versions
of python-pam. Also, this version avoids monkeypatching the PAM module.


liZe:
Thank you!

PR #284, opened 2015-05-04
Use path for git commit message

mmonaco:
The hard-coded message isn't very helpful, and that info is shown by the
author when looking at single line logs.


liZe:
Thank you!

PR #286, opened 2015-05-06
Improve Git commits

dnnr:
This PR is a resurrection and extension of the outdated #183.

It changes replace() to be actually implemented instead of composing remove() and append(), so that it performs not two independent but a single, concise transaction on the storage backend. This is generally more efficient and especially desirable when Git support is enabled.

It also makes the commit messages slightly more informative and reduces the amount of changes in each commit by sorting fields and removing PRODID beforehand.

### DIFF
--- a/radicale/ical.py
+++ b/radicale/ical.py
@@ -35,11 +35,14 @@ from contextlib import contextmanager
 import log
 
 
-def serialize(tag, headers=(), items=()):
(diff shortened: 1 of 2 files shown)


liZe:
Git support has been removed from Radicale and will be possible using external tools (see #372). Thank you!

PR #290, opened 2015-06-04
Handle read-only permissions in PROPFIND request

hadleyrich:
I'm using Davdroid and wanted to add a couple of read-only calendars.

I have no idea if this is a good way to implement this - I have only had a basic look through the code flow. If there's a better way, please do let me know.

That said, it works for me in my limited testing. I've only tested with Davdroid, Thunderbird, and Evolution. Davdroid and Evolution detect the calendar as read-only, Thunderbird does not.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -542,9 +542,8 @@ class Application(object):
         headers = {
             "DAV": "1, 2, 3, calendar-access, addressbook, extended-mkcol",
             "Content-Type": "text/xml"}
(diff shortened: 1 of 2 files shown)


madalu:
I applied this patch and it works well for me. This would be a great feature to have in radicale for users of davdroid (the best F-Droid caldav sync app for Android). Davdroid only makes a calendar read-only if the server announces it.

liZe:
Thank you. It was already fixed by #363, but I finally prefer your version.

Issue #293, opened 2015-07-16
Support the htpasswd md5 method


jgehrcke:
In the docstring of Radicale's htpasswd module we read 

> Plain-text, crypt and sha1 are supported, but md5 is not (see ``htpasswd`` man page to understand why).

I have grasped through the man page, have read through http://httpd.apache.org/docs/2.4/programs/htpasswd.html and through http://stackoverflow.com/questions/4175707/adding-a-salt-to-htpasswd and still ask myself: why would the md5 (htpasswd's default method) not be supported? The advantage over the sha1 method clearly would be that a random salt is used. The advantage over the crypt method would be that it is "more secure" (longer salt, password not limited to 8 chars, larger number of rounds).

If the reason is that the named md5 method has no equivalent in Python's stdlib (as it is "a version  of MD5 modified for Apache") then I believe that this should not necessarily be a deal breaker. The method is documented and has been implemented by e.g. passlib, too. Documentation: https://pythonhosted.org/passlib/lib/passlib.hash.apr_md5_crypt.html
Implementation: https://code.google.com/p/passlib/source/browse/passlib/handlers/md5_crypt.py

In view of the existing Python implementation: What do you think -- does it make sense for Radicale to support htpasswd's md5 method?

Btw: passlib uses the BSD license.

liZe:
Yes, it does, as long as other methods can still be used without passlib installed.

jgehrcke:
Okay, I'll look into providing a patch.

daks:
just my 2 cents: why still use MD5 method to hash passwords? MD5 is really old and subject to a lot of security problems.

jgehrcke:
This is mainly about supporting the different methods that Apache's htpasswd program provides for storing credentials. See httpd.apache.org/docs/programs/htpasswd.html. Whoever deploys Radicale has to make a deliberate choice about which encryption scheme to use.

BCRYPT and APR1 MD5 are, in this order, currently the most secure password encryption schemes that htpasswd provides. APR1 MD5 is supported by all versions of htpasswd, whereas BCRYPT is only supported since versions 2.4.x.

The MD5 method we are talking about here is "a version of MD5 modified for Apache", also called "APR1". It actually uses a random salt and many computation rounds. That is, this is highly different from simple canonical MD5 hashing in the sense that rainbow table attacks are more difficult to perform, and brute-forcing requires a certain amount of computational power.

This APR1 MD5 method is 
* the default password encryption scheme of the htpasswd program.
[...]

PR #297, opened 2015-07-24
htpasswd-based authentication: implement MD5-APR1 and BCRYPT support via passlib

jgehrcke:
This addresses https://github.com/Kozea/Radicale/issues/293.

### DIFF
--- a/radicale/auth/htpasswd.py
+++ b/radicale/auth/htpasswd.py
@@ -19,17 +19,43 @@
 # along with Radicale.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Htpasswd authentication.
+Implement htpasswd authentication.
 
-Load the list of login/password couples according a the configuration file
-created by Apache ``htpasswd`` command. Plain-text, crypt and sha1 are
-supported, but md5 is not (see ``htpasswd`` man page to understand why).
+Apache's htpasswd command (httpd.apache.org/docs/programs/htpasswd.html) manages
+a file for storing user credentials. It can encrypt passwords using different
+methods, e.g. BCRYPT, MD5-APR1 (a version of MD5 modified for Apache), SHA1, or
+by using the system's CRYPT routine. The CRYPT and SHA1 encryption methods
+implemented by htpasswd are considered as insecure. MD5-APR1 provides medium
+security as of 2015. Only BCRYPT can be considered secure by current standards.
 
+MD5-APR1-encrypted credentials can be written by all versions of htpasswd (its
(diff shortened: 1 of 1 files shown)


jgehrcke:
Not really sure what happened: this should be just two commits, not three (the empty merge is weird). Locally I had rebased onto master and squashed cleanly. Will look into this tomorrow, or is this already fine enough?

jgehrcke:
I have properly rebased my branch, so this could now be merged w/o polluting history. No we can discuss this (feedback really appreciated).

I think this should necessarily wait for Radicale 2.0. I believe it should be merged into the latest release line, because it enables **secure** password storage via htpasswd for the first time. This would also close the one-year-old discussion in #118. There, usage of passlib was also suggested. I have seen in multiple instances on the web that people mess with Radicale's htpasswd code in production for "fixing" the lack of proper password encryption themselves. This would not be necessary anymore. This is a feature that users desperately need, maybe without knowing.

### Overview 
The implementation proposed in this pull request implements a "progressive enhancement", i.e. provides additional crypto features upon availability of passlib and upon availability of a bcrypt backend for passlib. It provides meaningful error messages in case a certain crypto method is requested via config file, but is not supported by the run time.

### Error messages
The error messages provided by the proposed implementation (about mismatch between config and runtime availability) actually are exception values within RuntimeError exceptions. I have looked at the current architecture of Radicale and figured that this probably fits the existing code best (an alternative would have been to error out via `sys.exit("error message")`, for example). Feedback appreciated. It leaves to be mentioned that named mismatch is detected upon Radicale *start up*, instead of only catching this upon the first request.

### Tests
Clearly my goal was to not touch or invalidate existing unit tests. Didn't do so. However, the test coverage is not too great so far. I think proper testing of all htpasswd features proposed in this implementation requires some better unit test infrastructure, but I did not want to go ahead and propose drastic changes in this regard.

So, most importantly, the current unit tests pass with the proposed changes and I have on top of that tested various scenarios manually. That is, I am quite certain that the proposed changes could be releasex as 0.X w/o breaking anyone's system when upgrading.

### Config file annotation
Once we decide to merge this, the config file commentary should also be updated to reflect the new features, but also their security implications. The config file should express in simple terms that certain methods are plain insecure to use. I can do so if desired.

Thanks for all your work on Radicale and I hope we can include these features!

liZe:
Really good job! I'm OK with the implementation choices, the code is clean and simple. Adding some tests and trying the code in real life are the next steps, and it will happen soon as it's no merged in master.

jgehrcke:
Thanks for the quick feedback!

PR #341, opened 2015-12-23
Prevent "regex injection"

Unrud:
If an attacker is able to authenticate with a user name like .* he can bypass limitations imposed by "owner_write" and "owner_only".


liZe:
Thank you!

PR #342, opened 2015-12-24
Introduce naming scheme for request handlers

Unrud:
The do_ prefix and upper case name allows easy
distinction between methods that handle requests
and other methods.
Without this distinction an attacker could
call arbitrary methods.
Currently there is no method that matches the
argument count, but that's easy to miss when new
methods are added.


liZe:
Thank you again!

PR #343, opened 2015-12-24
Secure path handling

Unrud:
This fixes a number of issues with dodgy path handling:

  * Paths like ``..``, ``../..`` or ``//`` are not sanitized correctly
  * The program crashes if a path doesn't start with ``base_prefix`` instead of showing an error message
  * On MS Windows the filesystem backend allows access to the first level of files on a drive.
    
    e.g. the path ``/c:/filename/dummy`` gets converted to ``c:\filename``
  * The multifilesystem backend allows access to arbitrary files on all platforms.

    e.g. set the UID of a ics file to ``/dev/null``

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -32,7 +32,6 @@ import os
 import sys
 import pprint
 import base64
-import posixpath
 import socket
 import ssl
 import wsgiref.simple_server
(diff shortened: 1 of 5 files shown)


liZe:
That's a serious one.

Do you have other security-related issues coming? I'll have to release a new version after that.

Thank you again!

Unrud:
No, I've not.

PR #349, opened 2016-01-13
New parameter "user_strip_at" to remove @domain from user.

prlw1:
Given a user who configures some clients using a userid and others using an email address, probably because of the clients' differing user interfaces, we found it simplest just to strip the domain from the email address to just use the userid (as they matched) as per this patch.

Comments?

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -280,6 +280,11 @@ class Application(object):
             user = environ.get("REMOTE_USER")
             password = None
 
+        if user is not None and config.getboolean("auth", "user_strip_at"):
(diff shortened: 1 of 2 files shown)


liZe:
@prlw1 Thank you for this pull request.

I understand your needs, but I'm not sure that I want to merge this feature as it is really specific. Now that we have a way to cleanly inherit from the `Auth` class, it should be really easy for you to create a custom auth module (where you strip the domain), to inherit from the auth module you need.

What do you think of that?

liZe:
We can safely implement this feature in an auth module now thanks to #500.

PR #351, opened 2016-01-15
Quick fix for multifilesystem

Unrud:


radicale/storage/multifilesystem.py
```diff
@@ -80,7 +82,7 @@ def remove(self, name):
     def text(self):
         components = (
             ical.Timezone, ical.Event, ical.Todo, ical.Journal, ical.Card)
-        items = set()
+        items = {}
```
liZe: Does it work with the `items.update()` line 96?

Unrud: ```self._parse(fd.read(), components)``` returns a ```dict``` with names of components as keys and components as values.
It was failing before because ```items.update```added the component names to the ```set``` instead of the components itself.
Alternatively to 5081fcbcd1f558f92341a04917a7984650f5d117 you can also change line 96 to something like ```items.update(self._parse(fd.read(), components).values())```

liZe: Oh, OK, the docstring of `_parse` was wrong about what was returned, I've fixed it.

liZe:
Thank you!

PR #353, opened 2016-01-27
Integrated server without busy waiting

Unrud:
Use select syscall to detect incoming connections on any of the servers and program shutdown.

Busy waiting is still used with python < 3.5 on MS Windows because the function socket.socketpair() is not available. It's possible to work around this by implementing this function yourselves (copying it from python3.5) but I don't know if it's worth the hassle.

### DIFF
--- a/radicale/__main__.py
+++ b/radicale/__main__.py
@@ -28,8 +28,9 @@ import atexit
 import os
 import sys
 import optparse
+import select
 import signal
-import threading
+import socket
 from wsgiref.simple_server import make_server
 
 from . import (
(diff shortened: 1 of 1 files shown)


Unrud:
I did some measurements with and without this PR. The server was polled every 15 Minutes by one client.

**Before**

    real    217m48.530s
    user    0m40.010s
    sys     1m10.520s

**After**

    real    220m19.656s
    user    0m0.850s
    sys     0m0.320s

That's a reduction in cpu time by a factor of about 100.

liZe:
> That's a reduction in cpu time by a factor of about 100.

Awesome! :clap: :tada:. Thank you!

I merge the pull request as soon as I'll start working on #364.

PR #357, opened 2016-02-13
Create the missing tables if necessary

### LINKED ISSUE #356: Database tables are missing in the first usage
When using **sqlite** for storage, there is an error in the first usage:

```
OperationalError: (sqlite3.OperationalError) no such table: item [SQL: u'SELECT count(*) AS count_1 \nFROM (SELECT item.name AS item_name, item.tag AS item_tag, item.collection_path AS item_collection_path \nFROM item \nWHERE item.collection_path = ?) AS anon_1'] [parameters: ('user',)]
```

The reason is obvious: in the first usage, the required tables have not been created yet.

As far as I can see, there are two solutions:

1. Use the command-line utility `sqilte3` (provided by **sqlite**):

        $ sqlite3 radicale.db < schema.sql

2. Use the `create_all()` method of **SQLAlchemy** to create tables automatically, and I think this is a better solution

RussellLuo:
Fixed #356.

### DIFF
--- a/radicale/storage/database.py
+++ b/radicale/storage/database.py
@@ -36,7 +36,8 @@ from .. import config, ical
 # pylint: disable=C0103
 Base = declarative_base()
 Session = sessionmaker()
-Session.configure(bind=create_engine(config.get("storage", "database_url")))
+engine = create_engine(config.get("storage", "database_url"))
+Session.configure(bind=engine)
 # pylint: enable=C0103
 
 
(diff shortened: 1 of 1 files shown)


liZe:
Thanks for your PR.

As it's defined in #372, the storage modules have been removed from Radicale and will be available thanks to external "plugins". Would you be interested in developing a database plugin and/or maintaining it? You can add your comments in #391.

PR #369, opened 2016-04-02
Add umask configuration option

sherter:
The possibility to configure the `umask` for files created by Radicale makes ugly wrapper scripts à la `umask xyz && radicale` obsolete.

Also, in daemon-mode it was previously not possible to influence the umask (it's [hardcoded to 0](https://github.com/sherter/Radicale/blob/87aec8c29d30cbcbd1c18dd73684fe813f2d5ece/radicale/__main__.py#L118)).

### DIFF
--- a/config
+++ b/config
@@ -165,6 +165,9 @@
 # Folder for storing local collections, created if not present
 #filesystem_folder = ~/.config/radicale/collections
 
+# Umask for files created by Radicale
+#umask =
+
 # Database URL for SQLAlchemy
 # dialect+driver://user:password@host/dbname[?key=value..]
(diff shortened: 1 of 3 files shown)


Unrud:
> ugly wrapper scripts à la ``umask xyz && radicale`` obsolete.

All modern init systems support setting umask, I think it's better to use them instead of scripts or replicating that functionality in Radicale.

> Also, in daemon-mode it was previously not possible to influence the umask

That functionality is also included in init systems, daemon-mode is not necessary.

liZe:
Thank you for this pull request.

As proposed by @Unrud, I think that we should rely on init systems.

PR #371, opened 2016-04-02
Fix collection discovery in multifilesystem

Unrud:


liZe:
Thank you! It's been fixed by the API change.

PR #376, opened 2016-04-13
Various fixes

untitaker:
The vdirsyncer testsuite still fails, a lot of etag mismatches. Will look into
it some other time.


liZe:
Thank you!

PR #379, opened 2016-04-19
Fix double slash in REPORT responses

untitaker:
When doing REPORTs with absolute paths as hrefs, the response hrefs
contain a double slash at the beginning. This breaks URL parsers and
makes them assume they have a URL without protocol of the format
`//example.com/foo/bar/`.

I'd rather replace the entire statement with `uri = hreference` as
prepending the doubleslash seems unnecessary, but I haven't tested with
other request input.

### DIFF
--- a/radicale/xmlutils.py
+++ b/radicale/xmlutils.py
@@ -565,7 +565,7 @@ def report(path, xml_request, collection):
             # TODO: fix this
             if hreference.split("/")[-1] == item.href:
                 # Happening when depth is 0
-                uri = "/" + hreference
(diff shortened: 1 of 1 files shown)


liZe:
> I'd rather replace the entire statement with uri = hreference

You can. I suppose that this bug has been introduced by https://github.com/Kozea/Radicale/commit/bf8bd90bcba006d4f7bafed912d97d36ef23b637

untitaker:
Done.

liZe:
Thank you!

PR #382, opened 2016-04-20
Unbreak tag filters

untitaker:


untitaker:
Regarding implementing nested tag filters: I actually can't imagine a valid iCalendar construct where the current implementation wouldn't work. We might need a proper implementation for prop-filters though.

PR #385, opened 2016-04-22
Fixes to Rights

untitaker:
- BaseRights expects the config and logger params
- user may be None, which leads to errors when trying to use it with
  regexes.


liZe:
Thank you!

PR #397, opened 2016-05-18
Use tox within travis

untitaker:
This deduplicates the testing setup and commands


liZe:
Good idea. I'll use pytest in tox too, as nosetest is now deprecated.

liZe:
I've fixed many little things to fix that, there are lots of duplicates in tox.ini and .travis.yml, I'm not really happy with that…

PR #399, opened 2016-05-18
Fix tox env once and for all

untitaker:


untitaker:
See https://github.com/Kozea/Radicale/pull/397

liZe:
Thanks a lot!

Issue #388, opened 2016-05-02
Radicale stops working after some time


konmax:
Hi,

I use the latest version of Radicale (1.1.1) on a Win 7 Prof box and it works great except that it stops working after some time. I use it in standalone mode without a real web server in front of it and I know that the python web server has some issues and stops processing requests after some time. In the past, this happened from time to time and was never a big deal for me. However, something seem the have changed and now Radicale stops working after 1 or 2 days and this is really annoying. Whenever this happens, I can see this output in the console window:

```
C:\xxx\Radicale>set RADICALE_CONFIG=c:\xxx\Radicale\config

C:\xxx\Radicale>title Radicale CalDav Server

C:\xxx\Radicale>cd c:\xxx\Radicale

C:\xxx\Radicale>set PYTHONIOENCODING=utf-8

C:\xxx\Radicale>"C:\Program Files (x86)\Python34\python.exe" radicale.py
Traceback (most recent call last):
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 138, in run
    self.finish_response()
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 180, in finish_response
    self.write(data)
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 279, in write
    self._write(data)
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 453, in _write
    self.stdout.write(data)
  File "C:\Program Files (x86)\Python34\lib\socket.py", line 394, in write
    return self._sock.send(b)
  File "C:\Program Files (x86)\Python34\lib\ssl.py", line 684, in send
    v = self._sslobj.write(data)
ConnectionResetError: [WinError 10054] Eine vorhandene Verbindung wurde vom Remotehost geschlossen
----------------------------------------
Exception happened during processing of request from ('178.26.196.8', 4813)
Traceback (most recent call last):
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 138, in run
    self.finish_response()
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 180, in finish_response
    self.write(data)
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 279, in write
    self._write(data)
  File "C:\Program Files (x86)\Python34\lib\wsgiref\handlers.py", line 453, in _write
    self.stdout.write(data)
[...]

Issue #130, opened 2014-03-19
Discussion: Proposed changes to storage api for 1.0


jdccdevel:
I'm writing a custom storage backend, and the more I dig into the storage API, the more I see opportunities for improved efficiency.

This seems like a reasonable forum to discuss those improvements.

For example, several places in the code, I've seen this:
"if name in (item.name for item in collection.items)"

I'd like to start by suggesting a "has_item(name)" method be added to the "Collection" object to replace this. Alternatively, an "item_names" property could be created, but I believe that would be less efficient. 

The code right now is REALLY inefficient, as it requires the storage backend to load the ENTIRE collection, just to test if the collection contains a particular item name. This doesn't allow for any indexing or caching of item names, and hinders any efficiency improvements possible with databases.

This code pattern is repeated with the headers and properties, but it's less of an issue there due to their relatively small size.

liZe:
@untitaker your API is wonderful, thanks a lot!

PR #402, opened 2016-05-20
Implement locking of whole storage

Unrud:
Added method ``acquire_lock`` to ``BaseCollection`` which locks the whole storage for shared or exclusive access. In the default backend file locking is used to implement this method.
Unfortunately file locking is a bit broken and doesn't work in some cases, see comments.
This is not the most efficient method (all requests which potentially write to the storage are serialized) but it's simple.
For higher performance an transaction based approach would be great but that's difficult to implement on a raw filesystem and a database backend should be used.

I also enabled threading in the integrated webserver, this should resolve issues like #388.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -282,45 +282,37 @@ class Application:
         is_authenticated = self.is_authenticated(user, password)
         is_valid_user = is_authenticated or not user
 
-        lock = None
-        try:
-            if is_valid_user:
-                if function in (self.do_GET, self.do_HEAD,
-                                self.do_OPTIONS, self.do_PROPFIND,
-                                self.do_REPORT):
-                    lock_mode = "r"
-                else:
-                    lock_mode = "w"
-                lock = self.Collection.acquire_lock(lock_mode)
+        # Get content
+        content_length = int(environ.get("CONTENT_LENGTH") or 0)
+        if content_length:
+            content = self.decode(
+                environ["wsgi.input"].read(content_length), environ)
+            self.logger.debug("Request content:\n%s" % content)
+        else:
+            content = None
 
+        if is_valid_user:
+            if function in (self.do_GET, self.do_HEAD,
+                            self.do_OPTIONS, self.do_PROPFIND,
+                            self.do_REPORT):
+                lock_mode = "r"
+            else:
+                lock_mode = "w"
+            with self.Collection.acquire_lock(lock_mode):
                 items = self.Collection.discover(
                     path, environ.get("HTTP_DEPTH", "0"))
                 read_allowed_items, write_allowed_items = (
                     self.collect_allowed_items(items, user))
-            else:
-                read_allowed_items, write_allowed_items = None, None
-
-            # Get content
-            content_length = int(environ.get("CONTENT_LENGTH") or 0)
-            if content_length:
-                content = self.decode(
-                    environ["wsgi.input"].read(content_length), environ)
-                self.logger.debug("Request content:\n%s" % content)
-            else:
-                content = None
-
-            if is_valid_user and (
-                    (read_allowed_items or write_allowed_items) or
-                    (is_authenticated and function == self.do_PROPFIND) or
-                    function == self.do_OPTIONS):
-                status, headers, answer = function(
-                    environ, read_allowed_items, write_allowed_items, content,
-                    user)
-            else:
-                status, headers, answer = NOT_ALLOWED
-        finally:
-            if lock:
-                lock.release()
+                if (read_allowed_items or write_allowed_items or
+                        is_authenticated and function == self.do_PROPFIND or
+                        function == self.do_OPTIONS):
+                    status, headers, answer = function(
+                        environ, read_allowed_items, write_allowed_items,
+                        content, user)
+                else:
+                    status, headers, answer = NOT_ALLOWED
+        else:
+            status, headers, answer = NOT_ALLOWED
 
         if (status, headers, answer) == NOT_ALLOWED and not is_authenticated:
             # Unknown or unauthorized user
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -30,6 +30,7 @@ import os
 import pprint
 import base64
 import socket
+import socketserver
 import ssl
 import wsgiref.simple_server
 import re
@@ -90,6 +91,14 @@ class HTTPSServer(HTTPServer):
 
         self.server_bind()
         self.server_activate()
+
+
+class ThreadedHTTPServer(socketserver.ThreadingMixIn, HTTPServer):
+    pass
+
+
+class ThreadedHTTPSServer(socketserver.ThreadingMixIn, HTTPSServer):
+    pass
 
 
 class RequestHandler(wsgiref.simple_server.WSGIRequestHandler):
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -273,32 +273,45 @@ class Application:
         is_authenticated = self.is_authenticated(user, password)
         is_valid_user = is_authenticated or not user
 
-        if is_valid_user:
-            items = self.Collection.discover(
-                path, environ.get("HTTP_DEPTH", "0"))
-            read_allowed_items, write_allowed_items = (
-                self.collect_allowed_items(items, user))
-        else:
-            read_allowed_items, write_allowed_items = None, None
+        lock = None
+        try:
+            if is_valid_user:
+                if function in (self.do_GET, self.do_HEAD,
+                                self.do_OPTIONS, self.do_PROPFIND,
+                                self.do_REPORT):
+                    lock_mode = "r"
+                else:
+                    lock_mode = "w"
+                lock = self.Collection.acquire_lock(lock_mode)
 
-        # Get content
-        content_length = int(environ.get("CONTENT_LENGTH") or 0)
-        if content_length:
-            content = self.decode(
-                environ["wsgi.input"].read(content_length), environ)
-            self.logger.debug("Request content:\n%s" % content)
-        else:
-            content = None
+                items = self.Collection.discover(
+                    path, environ.get("HTTP_DEPTH", "0"))
+                read_allowed_items, write_allowed_items = (
+                    self.collect_allowed_items(items, user))
+            else:
+                read_allowed_items, write_allowed_items = None, None
 
-        if is_valid_user and (
-                (read_allowed_items or write_allowed_items) or
-                (is_authenticated and function == self.do_PROPFIND) or
-                function == self.do_OPTIONS):
-            status, headers, answer = function(
-                environ, read_allowed_items, write_allowed_items, content,
-                user)
-        else:
-            status, headers, answer = NOT_ALLOWED
+            # Get content
+            content_length = int(environ.get("CONTENT_LENGTH") or 0)
+            if content_length:
+                content = self.decode(
+                    environ["wsgi.input"].read(content_length), environ)
+                self.logger.debug("Request content:\n%s" % content)
+            else:
+                content = None
+
+            if is_valid_user and (
+                    (read_allowed_items or write_allowed_items) or
+                    (is_authenticated and function == self.do_PROPFIND) or
+                    function == self.do_OPTIONS):
+                status, headers, answer = function(
+                    environ, read_allowed_items, write_allowed_items, content,
+                    user)
+            else:
+                status, headers, answer = NOT_ALLOWED
+        finally:
+            if lock:
+                lock.release()
 
         if (status, headers, answer) == NOT_ALLOWED and not is_authenticated:
             # Unknown or unauthorized user
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -277,13 +277,12 @@ class BaseCollection:
         raise NotImplementedError
 
     @classmethod
+    @contextmanager
     def acquire_lock(cls, mode):
-        """Lock the whole storage.
+        """Set a context manager to lock the whole storage.
 
         ``mode`` must either be "r" for shared access or "w" for exclusive
         access.
-
-        Returns an object which has a method ``release``.
 
         """
         raise NotImplementedError
@@ -521,6 +520,7 @@ class Collection(BaseCollection):
     _lock = threading.Lock()
 
     @classmethod
+    @contextmanager
     def acquire_lock(cls, mode):
         class Lock:
             def __init__(self, release_method):
@@ -574,4 +574,5 @@ class Collection(BaseCollection):
             # TODO: use readers–writer lock
             cls._lock.acquire()
             lock = Lock(cls._lock.release)
-        return lock
+        yield
+        lock.release()
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -65,6 +65,13 @@ if os.name == "nt":
                              ctypes.wintypes.DWORD,
                              ctypes.POINTER(Overlapped)]
     lock_file_ex.restype = ctypes.wintypes.BOOL
+    unlock_file_ex = ctypes.windll.kernel32.UnlockFileEx
+    unlock_file_ex.argtypes = [ctypes.wintypes.HANDLE,
+                               ctypes.wintypes.DWORD,
+                               ctypes.wintypes.DWORD,
+                               ctypes.wintypes.DWORD,
+                               ctypes.POINTER(Overlapped)]
+    unlock_file_ex.restype = ctypes.wintypes.BOOL
 elif os.name == "posix":
     import fcntl
 
@@ -519,6 +526,7 @@ class Collection(BaseCollection):
 
     _lock = threading.Condition()
     _lock_file = None
+    _lock_file_locked = False
     _readers = 0
     _writer = False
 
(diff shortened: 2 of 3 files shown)


radicale/storage.py
```diff
@@ -474,3 +517,61 @@ def serialize(self):
...
+        if locked:
+            lock = Lock(lock_file.close)
+        else:
+            cls.logger.debug("Locking not supported")
+            lock_file.close()
+            # Fallback to primitive lock which only works within one process
+            # and doesn't distinguish between shared and exclusive access.
+            # TODO: use readers–writer lock
+            cls._lock.acquire()
+            lock = Lock(cls._lock.release)
```
untitaker: You can do `lock = cls._lock`.

radicale/storage.py
```diff
@@ -474,3 +517,61 @@ def serialize(self):
         elif self.get_meta("tag") == "VADDRESSBOOK":
             return "".join([item.serialize() for item in items])
         return ""
+
+    _lock = threading.Lock()
+
+    @classmethod
+    def acquire_lock(cls, mode):
```
untitaker: You might want to experiment with making this entire method just a contextmanager, to avoid having to return a stub Lock instance.

radicale/storage.py
```diff
@@ -474,3 +517,61 @@ def serialize(self):
...
+            overlapped = Overlapped()
+            if lock_file_ex(handle, flags, 0, 1, 0, overlapped):
+                locked = True
+        elif os.name == "posix":
+            operation = fcntl.LOCK_EX if mode == "w" else fcntl.LOCK_SH
+            # According to documentation flock() is emulated with fcntl() on
+            # some platforms. fcntl() locks are not associated with an open
+            # file descriptor. The same file can be locked multiple times
+            # within the same process and if any fd of the file is closed,
+            # all locks are released.
```
untitaker: Python threads all run in the same process. Therefore all thread workers may acquire locks at any time, exclusive or shared. Am I missing something?

untitaker:
I'm still a bit confused about the internal APIs of Radicale, but as far as I understand, a new class is created for each root directory (collection `/`), and objects are instantiated for subpaths [1]. That means that, because `acquire_lock` is a classmethod, it's effectively impossible to implement "finer-grained locking", e.g. per-subpath.

[1] I am not sure if I like this. I'd prefer a pathlib-like design where the root path is just a Collection instance like subpaths.

untitaker:
BTW there's the [filelock](https://pypi.python.org/pypi/filelock/) library which doesn't implement RwLocks, but has a sane API that doesn't allow one process to acquire the lock multiple times.

radicale/storage.py
Unrud: No, flock() locks are associated with a file descriptor (and it's duplicates) and not with a process.
fcntl() or lockf() locks can be acquired at any time within the same process.

Unrud:
> That means that, because ``acquire_lock`` is a classmethod, it's effectively impossible to implement "finer-grained locking", e.g. per-subpath.

Yes, I wanted to keep it simple. It improves the current situation by allowing simultaneous read access and the internal webserver doesn't hang-up on rogue connections.
Implementing "finer-grained locking" is difficult with the current architecture, especially on raw filesystems on multiple platforms.
As I wrote before, you should use a database backend with support for transactions if you want higher performance. You could even use the ``acquire_lock`` and ``release`` methods to start and finish the transaction.

> BTW there's the filelock library

As you wrote yourself, it doesn't support shared locks. It uses the existence of plain files for locking as a fallback. They are obviously not deleted than the program or computer crashes and Radicale would stop working until they are manually deleted.

radicale/storage.py
Unrud: Good idea.

Unrud:
@untitaker Changed ``acquire_lock`` over to context manager.

Only one lock file per process is used now, which allows the usage of a single file descriptor and lockf().

Also added some provision to prevent the starvation of writers. It's probably not worth the hassle, especially since lockf() doesn't bother.

radicale/storage.py
untitaker: That was my point. Since all threads are in the same process, the lock is ineffective.

On 21 May 2016 22:45:32 CEST, Unrud <notifications@github.com> wrote:
>> +        except OSError:
>> +            cls.logger.debug("Failed to set permissions on lock
>file")
>> +        locked = False
>> +        if os.name == "nt":
>> +            handle = msvcrt.get_osfhandle(lock_file.fileno())
>> +            flags = LOCKFILE_EXCLUSIVE_LOCK if mode == "w" else 0
>> +            overlapped = Overlapped()
>> +            if lock_file_ex(handle, flags, 0, 1, 0, overlapped):
>> +                locked = True
>> +        elif os.name == "posix":
>> +            operation = fcntl.LOCK_EX if mode == "w" else
>fcntl.LOCK_SH
>> +            # According to documentation flock() is emulated with
>fcntl() on
>> +            # some platforms. fcntl() locks are not associated with
>an open
>> +            # file descriptor. The same file can be locked multiple
>times
>> +            # within the same process and if any fd of the file is
>closed,
>> +            # all locks are released.
>
>No, flock() locks are associated with a file descriptor (and it's
>duplicates) and not with a process.
>fcntl() or lockf() locks can be acquired at any time within the same
>process.
>
>---
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/Kozea/Radicale/pull/402/files/ff3b31fed6d1c090e4957a4c3d9432a375f2d216#r64140033

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Unrud: I think you misunderstood me. flock() and lockf() locks behave very different. Everything you said only applies to lockf() locks. flock() locks, which are used here, are not ineffective.

untitaker: On line 594 you do call lockf though.

On 23 May 2016 09:36:35 CEST, Unrud <notifications@github.com> wrote:
>> +        except OSError:
>> +            cls.logger.debug("Failed to set permissions on lock
>file")
>> +        locked = False
>> +        if os.name == "nt":
>> +            handle = msvcrt.get_osfhandle(lock_file.fileno())
>> +            flags = LOCKFILE_EXCLUSIVE_LOCK if mode == "w" else 0
>> +            overlapped = Overlapped()
>> +            if lock_file_ex(handle, flags, 0, 1, 0, overlapped):
>> +                locked = True
>> +        elif os.name == "posix":
>> +            operation = fcntl.LOCK_EX if mode == "w" else
>fcntl.LOCK_SH
>> +            # According to documentation flock() is emulated with
>fcntl() on
>> +            # some platforms. fcntl() locks are not associated with
>an open
>> +            # file descriptor. The same file can be locked multiple
>times
>> +            # within the same process and if any fd of the file is
>closed,
>> +            # all locks are released.
>
>I think you misunderstood me. flock() and lockf() locks behave very
>different. Everything you said only applies to lockf() locks. flock()
>locks, which are used here, are not ineffective.
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Kozea/Radicale/pull/402/files/ff3b31fed6d1c090e4957a4c3d9432a375f2d216#r64179106

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

untitaker: Also from the manpage flock(2):

>blocking if an incompatible lock is held by another process

There is no indication it would block if another thread holds the lock.

On 23 May 2016 09:36:35 CEST, Unrud <notifications@github.com> wrote:
>> +        except OSError:
>> +            cls.logger.debug("Failed to set permissions on lock
>file")
>> +        locked = False
>> +        if os.name == "nt":
>> +            handle = msvcrt.get_osfhandle(lock_file.fileno())
>> +            flags = LOCKFILE_EXCLUSIVE_LOCK if mode == "w" else 0
>> +            overlapped = Overlapped()
>> +            if lock_file_ex(handle, flags, 0, 1, 0, overlapped):
>> +                locked = True
>> +        elif os.name == "posix":
>> +            operation = fcntl.LOCK_EX if mode == "w" else
>fcntl.LOCK_SH
>> +            # According to documentation flock() is emulated with
>fcntl() on
>> +            # some platforms. fcntl() locks are not associated with
>an open
>> +            # file descriptor. The same file can be locked multiple
>times
>> +            # within the same process and if any fd of the file is
>closed,
>> +            # all locks are released.
>
>I think you misunderstood me. flock() and lockf() locks behave very
>different. Everything you said only applies to lockf() locks. flock()
>locks, which are used here, are not ineffective.
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Kozea/Radicale/pull/402/files/ff3b31fed6d1c090e4957a4c3d9432a375f2d216#r64179106

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

untitaker: Also the Linux manpage of fcntl doesn't indicate any different behavior: http://linux.die.net/man/2/fcntl

On 23 May 2016 09:36:35 CEST, Unrud <notifications@github.com> wrote:
>> +        except OSError:
>> +            cls.logger.debug("Failed to set permissions on lock
>file")
>> +        locked = False
>> +        if os.name == "nt":
>> +            handle = msvcrt.get_osfhandle(lock_file.fileno())
>> +            flags = LOCKFILE_EXCLUSIVE_LOCK if mode == "w" else 0
>> +            overlapped = Overlapped()
>> +            if lock_file_ex(handle, flags, 0, 1, 0, overlapped):
>> +                locked = True
>> +        elif os.name == "posix":
>> +            operation = fcntl.LOCK_EX if mode == "w" else
>fcntl.LOCK_SH
>> +            # According to documentation flock() is emulated with
>fcntl() on
>> +            # some platforms. fcntl() locks are not associated with
>an open
>> +            # file descriptor. The same file can be locked multiple
>times
>> +            # within the same process and if any fd of the file is
>closed,
>> +            # all locks are released.
>
>I think you misunderstood me. flock() and lockf() locks behave very
>different. Everything you said only applies to lockf() locks. flock()
>locks, which are used here, are not ineffective.
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Kozea/Radicale/pull/402/files/ff3b31fed6d1c090e4957a4c3d9432a375f2d216#r64179106

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Unrud: 1. Yeah, but that's a newer version of the code.
Only one lock file is used there, see 49bc0728e3c2dc0d57f73e0d7af33bc9cda65e0e
 2. The manpage also states:
    
        Locks created by flock() are associated with an open file description
    
    But it doesn't really matter as lockf() is used now anyway.

Unrud: 3. The fcntl manpage states:
    
        File locks are released as soon as the process holding the locks closes some file descriptor for  the  file.
    
    This indicates that the locks are associated with a process and not with an open file description. But both manpages could be written clearer.

untitaker:
```py
import fcntl

fds = []
for _ in range(10):
    f = open('foo', 'w+')
    fds.append(f)

    fcntl.lockf(f.fileno(), fcntl.LOCK_EX)

    print("Locked")
```

This doesn't crash and prints "Locked" 10 times.

Unrud: > This doesn't crash and prints "Locked" 10 times.

I wrote that ``lockf`` (aka ``fcntl``) locks are associated with a process and not with an open file description, don't know what you are trying to prove here. Maybe you mixed up ``lockf`` and ``flock``?

liZe:
Seems to be really cool, thank you very much!

Just a couple of remarks/questions:
- The goal of `acquire_lock` is the same as the goal of the (not implemented) `at_once`, right? If it is, I'd prefer to use `at_once` to keep the same API as vdirsyncer (it's discussed in #130).
- I'd prefer to put `MAX_FILE_LOCK_DURATION` in a configuration key (I can do this later myself).
- I'm a bit frightened by the NT code, but I now understand why using filelock is probably not a solution.
- `ThreadedHTTPServer` is just perfect, it's a huge improvement.
- The performance side will be improved in the future, your current code is quite simple and I like it, it's probably better for now.

radicale/storage.py
untitaker: Then change the code to use a single file descriptor. It still exhibits the same behavior.

untitaker: And flock exhibits the same behavior anyway.

radicale/__init__.py
```diff
@@ -92,6 +93,14 @@ def __init__(self, address, handler):
         self.server_activate()
 
 
+class ThreadedHTTPServer(socketserver.ThreadingMixIn, HTTPServer):
```
untitaker: Unsure if daemon_thread = True should be set. If I blast the server with requests, it can't be quit properly.

untitaker: The concern I have with `daemon_thread = True` is that partial uploads get aborted.

Perhaps atomic file writes (write to temporary file and then move) is the solution.

radicale/storage.py
Unrud: > And flock exhibits the same behavior anyway.

No, at least not on Linux. "Locked" is only printed once and the program hangs. Maybe you are on one of the platforms the documentation mentions that use ``fcntl`` to emulate ``flock``.

That possible incompatibility is the reason I changed over to ``lockf`` in later commits.

> Then change the code to use a single file descriptor. It still exhibits the same behavior.

Don't know what you mean by that. If you call ``lockf`` multiple times on the same file description, you just change the lock mode.

untitaker: Sorry, somehow a lot of my tests were wrong, and I was massively confused about what you wrote earlier. I now see that you reuse the file descriptor between workers, I wasn't aware of that. I hope my future comments are a bit clearer now.

radicale/__init__.py
Unrud: Atomic file writes are not sufficient to solve the issue of partial uploads, because changes of multiple files might be needed to transfer from one valid state to another (e.g. moving of whole collections).
Solutions to this problem are quite  elaborate. If it's a requirement, just use a database that supports ACID.

untitaker:
* This is basically a roundabout implementation of a simple RwLock, with platform-specific code. We already accept third-party dependencies, so we might as well depend on https://testpypi.python.org/pypi/rwlock or do our own implementation.
* One of the big things people wanted to do is to use gunicorn with Radicale. This is still not safely possible with this implementation and that's totally fine -- but it still should be pointed out.

radicale/__init__.py
untitaker: I forgot that one can MOVE collections in WebDAV.

Unrud:
> The goal of acquire_lock is the same as the goal of the (not implemented) at_once, right?

I think so.

> I'm a bit frightened by the NT code

Maybe it's better to just use ``msvcrt.locking``. Parallel read access with multiple processes will be lost on Windows, but shared access within one process will still work. That's only a problem if someone is using a WSGI server that spawns multiple processes, which probably nobody does on Windows.

> I'd prefer to put ``MAX_FILE_LOCK_DURATION`` in a configuration key (I can do this later myself).

I might as well revert 6b1acd14d70b2e029aa56905518b3185a66d91c2. That feature just adds unnecessary complexity most of the time. ``fcntl`` doesn't seem to bother with possible starvation, so it might not even work if three or more processes are used.

> This is basically an implementation RwLock, with platform-specific code. We already accept third-party dependencies, so we might as well depend on https://testpypi.python.org/pypi/rwlock or do our own implementation.

It's a implementation of a platform-independend RWLock that interlocks with a splatform-specific file-based RWLock. The interlocking part makes the usage of external RWLock libraries less attractive, as that part still needs to be added and complexity probably wouldn't get reduced in comparison to the current implementation.

untitaker:
And again I'm wrong. Nice work.

liZe:
> The goal of acquire_lock is the same as the goal of the (not implemented) at_once, right?
> I think so.

Thanks. I'll rename the function.

> Maybe it's better to just use msvcrt.locking. Parallel read access with multiple processes will be lost on Windows, but shared access within one process will still work. That's only a problem if someone is using a WSGI server that spawns multiple processes, which probably nobody does on Windows.

I'm sooooo far from Windows that it's a bit hard for me to tell what's important or not. @untitaker any idea about this?

> I might as well revert 6b1acd1. That feature just adds unnecessary complexity most of the time. lockf doesn't seem to bother with possible starvation, so it might not even work if three or more processes are used. Couldn't find documentation on this.

I'll revert it.

If everything is OK for you, I can merge ASAP. Thanks *a lot* for the code and the reviews.

untitaker:
`at_once` is probably taken from vdirsyncer. It doesn't really make sense here because Radicale only supports this one storage type where it is a no-op.

liZe:
> at_once is probably taken from vdirsyncer.

Yes, of course! (it was in my last comment)

> It doesn't really make sense here because Radicale only supports this one storage type where it is a no-op.

As `acquire_lock` implemented in this pull request has the same function as `at_once` in vdirsyncer, the proposition was to rename `acquire_lock` into `at_once`.

untitaker:
at_once in vdirsyncer is intended to be more of a performance optimization, but I guess the name is also appropriate for "transactions". I personally think "acquire_lock" is more appropriate.

liZe:
> I personally think "acquire_lock" is more appropriate.

OK. I'll remove `at_once` for now, as it probably won't be used before 2.0.

liZe:
:smile: :clap: :sunflower:

untitaker:
Just for completeness sake, I reran an old test script I had against Radicale: https://gist.github.com/untitaker/54239710c23c4a89dac03babd4e4ea06

Without this PR it shows data races.

PR #403, opened 2016-05-20
Don't sanitize local paths using posixpath

Unrud:
This doesn't work as expected on Windows.


liZe:
Thanks! I'll just add a line in the documentation to tell that `root` must be a clean filesystem path.

PR #414, opened 2016-05-27
Fix #413 about Apache access control directives

### LINKED ISSUE #413: Conflicting access control directives in Apache documentation example
In the documentation [here](http://radicale.org/user_documentation/#idapache-and-mod-wsgi), you have the example:

```apache
        AuthType Basic
        AuthName "Radicale Authentication"
        AuthBasicProvider file
        AuthUserFile /usr/share/radicale/radicale.passwd
        Require valid-user

        AllowOverride None
        Order allow,deny
        allow from all
```

There are two problems with this in Apache 2.4. The first is that having both `Require valid-user` and `Order allow,deny / allow from all` [mixes old and new directives](https://httpd.apache.org/docs/current/upgrading.html#access) and is discouraged. The second is that `Order allow,deny / allow from all` translates to `Require all granted`, and having two `Require` directives in the same configuration section means that if any of them pass, the request is allowed (see [here](https://httpd.apache.org/docs/2.4/mod/mod_authz_core.html#require) and [here](https://httpd.apache.org/docs/2.4/mod/mod_authz_core.html#requireany)), so a request here is always allowed and the `valid-user` restriction is irrelevant.

I think a simple fix is to just replace what you have with this instead:

```apache
        AuthType Basic
        AuthName "Radicale Authentication"
        AuthBasicProvider file
        AuthUserFile /usr/share/radicale/radicale.passwd

        AllowOverride None
        Require valid-user
```

jeremyn:


liZe:
Thank you!

PR #417, opened 2016-05-28
Allow creation of empty collection via PUT

Unrud:
If an empty collections is created with PUT, content is None.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -546,7 +546,7 @@ class Application:
             # Case 1: No item and no ETag precondition: Add new item
             # Case 2: Item and ETag precondition verified: Modify item
             # Case 3: Item and no Etag precondition: Force modifying item
-            items = list(vobject.readComponents(content))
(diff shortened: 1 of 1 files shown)


untitaker:
For creating collections you ought to use MKCOL/MKCALENDAR, not PUT. Am I missing something?

Also there probably should be some sort of assertion that only one item is uploaded, instead of silently discarding leftovers.

Unrud:
> For creating collections you ought to use MKCOL/MKCALENDAR, not PUT. Am I missing something?

I don't think that this is in the specification, but some clients use it.

untitaker:
Which clients in particular? I don't know of a server that supports this either, so I am not sure if it is possible.

Unrud:
> Which clients in particular?

Outlook

untitaker:
Ok, however your PR still doesn't seem to have this effect. PUT on collections returns 201 without doing anything.

liZe:
Thank you!

PR #418, opened 2016-05-28
Fix crash on unknown content type

Unrud:


untitaker:
I'd rather reject the request.

Unrud:
> I'd rather reject the request.

Outlook uses text/plain as mimetype when a new calendar is published via webdav.

untitaker:
I see!

liZe:
> Outlook uses text/plain as mimetype when a new calendar is published via webdav.

WHAT?

PR #419, opened 2016-05-29
Save all items with the same UID in the same file

Unrud:
- If recurrences are not in the same file, they are not correctly shown by clients.
  -  Try to use UID as filename

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -380,12 +380,25 @@ class Collection(BaseCollection):
                         for oitem in items[i+1:]:
                             if getattr(oitem, "uid", None) == uid:
                                 new_collection.add(oitem)
-                    self.upload(uuid4().hex, new_collection)
+                    if uid and is_safe_filesystem_path_component(uid.value):
+                        href = uid.value
+                    else:
+                        href = uuid4().hex
+                    if not href.lower().endswith(".ics"):
+                        href += ".ics"
+                    self.upload(href, new_collection)
         elif tag == "VCARD":
             self.set_meta("tag", "VADDRESSBOOK")
             if collection:
                 for card in collection:
-                    self.upload(uuid4().hex, card)
+                    uid = getattr(card, "uid", None)
+                    if uid and is_safe_filesystem_path_component(uid.value):
+                        href = uid.value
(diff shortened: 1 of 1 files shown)


liZe:
> Try to use UID as filename

I'm afraid of ghosts. I mean, I'm REALLY AFRAID of REALLY FRIGHTENING GHOSTS. No, no, don't tell me that mixing UIDs and filenames is going to find its way back into Radicale, these ghosts will hunt us in our nights and we'll NEVER be allowed to SLEEP AGAIN, **NEVER EVER AGAIN**, and they will kill our children and kittens with us, yes, ALL OF US, **ALL OF.

So. Is it OK for you to always use a UUID instead of trying the UID to create the filename? I think too that we don't need the ".ics" part. (And I hope that you only added UID and ".ics" in the filename for cosmetics :smile:.)

Unrud:
0b25c82d9d7b9bba874470ba439fb4bae08eeb1b is just cosmetics.
I don't see a difference between using the UID or some random UUID for the filenames, **when the files are first created.** Filenames and UIDs are not coupled in any way, the files don't get automatically renamed later on (I guess that was causing the problems before).

liZe:
> 0b25c82 is just cosmetics.

:heart: :tada:

I'll revert it if you don't mind, I'm superstitious.

Unrud:
> I'll revert it if you don't mind, I'm superstitious.

I'm okay with that.

liZe:
Thank you!

radicale/storage.py
```diff
@@ -364,17 +364,41 @@ def create_collection(cls, href, collection=None, tag=None):
...
+                for i, item in enumerate(items):
+                    uid = getattr(item, "uid", None)
+                    if uid in processed_uids:
+                        continue
+                    new_collection = vobject.iCalendar()
+                    new_collection.add(item)
+                    if uid:
+                        processed_uids.append(uid)
+                        # search for items with same UID
+                        for oitem in items[i+1:]:
```
untitaker: This is `O(n^2)`. It could be `O(n)` by having a dict that maps from UIDs to `iCalendar` objects.

PR #425, opened 2016-06-09
Don't silently drop files

Unrud:
The server should signal an error instead of silently dropping files/updates if an filesystem path is unsafe or an (secondary) ETag match fails.


untitaker:
Fixes #433

Unrud:
rebased

liZe:
Thank you!

PR #426, opened 2016-06-09
Ignore hidden files

Unrud:
Ignore files like ``.git``, ``.Radicale.lock`` or ``0d706cb6-5726-494f-b744-f8b8073105df~``.


liZe:
Why not, thanks!

Issue #372, opened 2016-04-04
Follow roadmap for 2.0


liZe:
As defined in #364.
- Stay simple:
  - Keep only what's needed:
    - [x] Remove Python 2.x support (#316).
    - [x] Keep only one simple filesystem-based storage system (#83, #114, #130, #134, #161, #206, #207, #215, #324, #357).
    - [x] Remove built-in Git support (#286, #317).
    - [x] Remove built-in authentication modules (#323).
  - Clean the documentation:
    - [x] Mention versions covered by the documentation (#157).
    - [x] Use GitHub to host the website (#310).
    - [ ] Create a new design for the website.
  - Write the documentation:
    - [x] Migration guide.
    - Install (#112)
      - [x] What is Radicale.
      - [x] Tutorial.
      - [x] Simple installation.
      - [x] Production-ready installation (#197, #266, #270, #276, #388).
      - [x] Versioning.
    - Use
      - [ ] Configuration of clients.
    - Configure
      - [x] What can be configured?
      - [x] Auth and rights. (#538)
      - [ ] Storage.
      - [x] Logging.
    - Hack
      - [x] How does Radicale work.
      - [x] Plugins.
      - [x] Debugging.
      - [x] Documentation.
- Become solid:
  - Use standards:
    - [x] Keep the WSGI interface, use Python HTTP server by default.
    - [x] Use a real iCal parser ([icalendar](http://icalendar.readthedocs.org/en/latest/), [vobject](http://eventable.github.io/vobject/)).
    - [x] Don't rely on slashes (#322).
    - [x] Build a solid calendar discovery (#292).
    - [x] Respect the difference between "files" and "folders" (#366).
    - [x] Remove the calendar creation with GET requests (#240, #248, #252).
    - [x] Define how to create collections (#384).
  - Be thread-safe:
    - [x] Be stateless (#122).
    - [x] Use a file locker ([filelock](https://pypi.python.org/pypi/filelock/)) (#14).
    - [x] Add threading (#266, #388).
    - [x] Get atomic writes (#440).
- Provide needed new features:
  - [x] Built-in filters (#33, #101).
  - [x] Read-only permissions (#290, #363).
  - [x] External authentication modules allowed (#187, #227, #247, #278, #325, #349).
  - [x] External rights modules allowed (#242, #374).
  - [x] External storage modules allowed (#130, #392).
  - [x] External version control tools allowed (like [gitwatch](https://github.com/nevik/gitwatch)) (#327).

[...]

Issue #327, opened 2015-10-25
Git hooks are not being executed


jreinert:
I added a post-commit git hook which checks if changes to the BDAY field of contacts in address books are made and generates a birthday calendar for this address book, adds and commits it. This hook isn't run though. I tried manually changing a BDAY field and committing it. The hook ran just fine. Maybe It's an issue with dulwich...

liZe:
According to its NEWS file, dulwich launches hooks as long as they're posix shell scripts. I've never used hooks with dulwich, you should ask its developers, it's probably possible to find a solution.

liZe:
Reported in dulwich: jelmer/dulwich#386.

liZe:
I close this ticket as the git support will not be built-in anymore.

Issue #289, opened 2015-05-18
Aggregate calendars


hobarrera:
I'd like to share my calendar(s) with a few people. However, I have several (personal, work, study, etc).

I'd like to share them all as a single URL (maybe `/path/to/calendars/*`?) (because, honestly, someone else wouldn't care about which calendar I put stuff in).  
I don't think this is a standard CalDAV feature, but I'd still love to see something like this on radicale.

This would be extra useful after fixing issue #34, since it'd be extra-useful to share all my calendars' free-busy as a single link.

Would something like this be doable? Would it be acceptable (since it's defined in any standard really)?

liZe:
As explained in #429, adding a hook seems to be a good way to do this.

PR #429, opened 2016-06-11
Add hook for storage changes

Unrud:
This allows easy integration of external version control tools. (#372, #327)

Example:
```
hook = git add -A && (git diff --cached --quiet || git commit -m "Changes by "%(user)s)
```

This is also a possible solution to #289:

Example:
```
hook = rm calendars/all/*; cp calendars/*/* calendars/all; true

### DIFF
--- a/config
+++ b/config
@@ -112,6 +112,10 @@
 # Folder for storing local collections, created if not present
 #filesystem_folder = ~/.config/radicale/collections
 
+# Command that is run after changes to storage
+#hook =
+# Example: git add -A && (git diff --cached --quiet || git commit -m "Changes by "%(user)s)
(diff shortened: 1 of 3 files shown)


untitaker:
Good idea, but should the request fail if the hook fails to execute?

Unrud:
> should the request fail if the hook fails to execute?

Otherwise it's difficult to detect errors. You would have to monitor logs or regularly check that your backup (or whatever you are using the hook for) is really working. This way your clients start to show errors if something is wrong.

untitaker:
Okay. LGTM!

liZe:
Thank you!

PR #436, opened 2016-06-26
Improve Dockerfile

Horgix:
This PR brings a lot of modifications to the existing Dockerfile, with the aim of improving it by making it more compliant to best practices and de-facto usages.
I know this PR contains huge modifications splitted in several commits; I've done this so it can be reviewed in an iterative "step by step" way, and am open to squashing it if you prefer to.

I'm also, of course, open to any suggestion/comment about it and don't want to force anything :)
I'll describe more in details the modifications below.

### Indentation/Alignment

The Dockerfile has been modified to make it more consistent with the official Dockerfiles (regarding purely style things such as tabs, alignment, etc), like for example the [nginx one](https://github.com/nginxinc/docker-nginx/blob/11fc019b2be3ad51ba5d097b1857a099c4056213/mainline/jessie/Dockerfile).

### Comments

I took the liberty in my rework to remove superfluous comments; a Dockerfile is, most of the time, self-explanatory.

## Directives

The Dockerfile was lacking a [MAINTAINER](https://docs.docker.com/engine/reference/builder/#maintainer) directive; I added one refering to the project and its mailing list.

I also added an [EXPOSE](https://docs.docker.com/engine/reference/builder/#expose) directive to make explicit the fact that it's listening on port 5232.

## Dependancies

- The dependancies handling part (apk and pip) have been merged in an unique layer
- apk commands have been merged as much as possible
- pip is installed from the dedicated built-in module of python3

## Sources

While directly COPY-ing sources is perfectly fine for projects delivering Docker images to the [Docker Hub](https://hub.docker.com/) with automated build, I think that in Radicale's case it would benefit from being more easily buildable by everyone.

The COPY of the sources have therefore been replaced by a direct download of the latest release from GitHub, which is then extracted and installed. Future releases will only need to update the "VERSION" variable.

## Radicale Configuration

With the same logic as sources, configuration is now directly taken from the sources already present in the image, and copied to a [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) compliant and [already documented](http://radicale.org/user_documentation/#idmain-configuration-file) path instead of `/data` : `/etc/radicale/config`

### -f

Even if it seems to be already the default behavior right now, I added the -f option to the command to make sure Radicale runs in foreground.

## Conclusion

I'm already using this reworked Dockerfile to run my own Radicale server and it's working fine.
If this PR gets accepted, I'm also willing to add a part about running Radicale in Docker in the documentation, if you're OK with this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,21 +17,10 @@ RUN apk --update --update-cache upgrade \
 
 RUN pip3 install passlib bcrypt
 
-RUN mkdir -p /data/config
 RUN wget ${TARBALL} \
     && tar xzf ${VERSION}.tar.gz \
     && cd Radicale-${VERSION} && python3 setup.py install
 
-COPY config /data/config
+COPY config /srv/radicale.conf
 
-# User
-RUN adduser -h /home/radicale -D radicale \
-      && mkdir -p /home/radicale/.config \
-      && ln -s /data/config /home/radicale/.config/radicale \
-      && chown -R radicale:radicale /data/config \
-      && chown -R radicale:radicale /home/radicale
-
-USER radicale
-WORKDIR /home/radicale
-
-CMD ["radicale", "-D", "-C", "/data/config/config"]
+CMD ["radicale", "-D", "-C", "/srv/radicale.conf"]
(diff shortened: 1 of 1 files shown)


liZe:
Thank you!

PR #444, opened 2016-07-14
Hide temporary files

### LINKED ISSUE #443: Temporary files from AtomicWriter as items
AtomicWriter creates temporary files in the form tmpXXX. If the program/computer crashes and the temporary file remains, it's loaded as an normal item the next time the collection is queried.

The temporary files should be named something like ``.Radicale.tmp-XXX`` so that they are ignored by the collection.

Unrud:
Fixes #443


liZe:
Thank you!

PR #445, opened 2016-07-14
Point well-known to /

### LINKED ISSUE #431: well-known targets wrong
``/.well-known/carddav`` points to ``/carddav/`` or ``'/%(user)s/carddav/'`` in the config file.

``/.well-known/caldav`` points to ``/caldav/`` or ``'/%(user)s/caldav/'`` in the config file.

I think that the well-known URLs should just point to the relevant  WebDAV server, ``/`` in our case.

Unrud:
Fixes #431

### DIFF
--- a/config
+++ b/config
@@ -73,10 +73,10 @@
 [well-known]
 
 # Path where /.well-known/caldav/ is redirected
-#caldav = '/%(user)s/caldav/'
+#caldav = /
 
 # Path where /.well-known/carddav/ is redirected
-#carddav = '/%(user)s/carddav/'
+#carddav = /
 
 
 [auth]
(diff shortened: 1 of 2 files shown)


liZe:
I have to check if the clients are happy with that before merging.

Unrud:
> I have to check if the clients are happy with that before merging.

According to the [specification](https://tools.ietf.org/html/rfc6764) clients query ``/.well-known/carddav``, then do a PROPFIND request on the redirection target looking for the ``DAV:current-user-principal`` property. If that fails, they check ``/`` for that property instead.

Currently ``/.well-known/carddav`` points to ``/cardav``, which does not exist and returns 404. The whole well-known thing seems pretty useless (at least for this application) and should probably be removed completely.

PR #450, opened 2016-08-01
Prevent unsafe usernames

Unrud:
If an attacker is able to create an account with a username like "user/calendar.ics", he can access collections of other users.


liZe:
Thank you!

PR #451, opened 2016-08-01
Remove /.well-known

Unrud:
I can't think of a sane use-case for this.

### DIFF
--- a/config
+++ b/config
@@ -70,15 +70,6 @@
 #stock = utf-8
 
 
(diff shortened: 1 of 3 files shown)


liZe:
This feature had been discussed in #126, but now that we redirect to the root path of the server, I agree with you, it's useless (and maybe a source of bugs). Let's remove it, we'll see if anybody has a sane use case.

PR #452, opened 2016-08-01
Always match full username/collection with regex

Unrud:
It's easy to forget $ at the end of a regex and it's counter-intuitive that ^ is implicit but $ is not.


liZe:
Good idea, thanks!

Issue #440, opened 2016-07-01
Data loss due to non-atomic writes


kentonv:
`storage.py` contains many clauses that update files by overwriting them, like:

    with open(path, "w", encoding=self.storage_encoding) as fd:
        fd.write(item.serialize())

Unfortunately, if a crash or machine failure occurs between the first line and the second line, the result is that the item will be completely lost, since `open()` erases all existing data and the data isn't replaced until the write.

Note that on a typical Linux system, there may be a full 30 seconds between the call to `write()` and the data actually being flushed to disk. Therefore, after this code, there is a 30-second window in which a power failure will cause data loss.

This is a real problem for people running Radicale on https://oasis.sandstorm.io: In such large distributed systems, "power failures" -- or events that look a lot like power failures, such as losing the network connection to the storage volume -- are common. We've now had two independent reports of Radicale losing people's calendars on Oasis.

To fix this problem, Radicale should always write data to a temporary file first, and then `os.rename()` the temporary over the original. Technically it should also do `os.fsync(f.fileno())` immediately before the rename (although on Linux / ext4 systems this is not strictly necessary).

untitaker:
python-atomicwrites handles this usecase although it might be overkill

liZe:
@untitaker I've JUST RIGHT NOW found your project too. What a timing.

liZe:
(Or you've just seen my last comment about this in #372, magic's not always magic.)

liZe:
I've used atomicwrites for now. It works, but I'd be happy to have real life tests.

Unrud:
This issue is not solved. ``AtomicWriter`` does not provide durability on non-windows systems (see untitaker/python-atomicwrites#12). ``Collection.delete`` also does not ensure that the changes are really written to disk.

It's still possible that deleted events reappear in clients and that created events or changes to existing events disappear.

kentonv:
To be clear:

* The fix as-implemented currently will ensure that people don't randomly lose their whole calendar unexpectedly, because the update is now atomic -- there's no point in time where a crash would lead to the file (both old and new versions) being totally lost.

* However, until you fsync() the directory, it is perhaps possible that the replacement operation hasn't been committed to disk. A crash at this point could mean that the modification is lost, which is bad if the user has been informed that the write succeeded. This is less disastrous than losing the whole calendar, but it's worth fixing as well.

untitaker:
[...]

PR #453, opened 2016-08-01
Delete atomic and durable

Unrud:
See #440

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -580,7 +580,21 @@ class Collection(BaseCollection):
         if href is None:
             # Delete the collection
             if os.path.isdir(self._filesystem_path):
-                shutil.rmtree(self._filesystem_path)
+                try:
+                    os.rmdir(self._filesystem_path)
+                except OSError:
+                    while True:
+                        tmp_filesystem_path = os.path.join(
+                            os.path.dirname(self._filesystem_path),
+                            ".Radicale.tmp-" + hex(getrandbits(32))[2:])
+                        if not os.path.exists(tmp_filesystem_path):
+                            break
+                    os.rename(self._filesystem_path, tmp_filesystem_path)
+                    sync_directory(os.path.dirname(self._filesystem_path))
+                    # Deferred because it might take a long time
+                    shutil.rmtree(tmp_filesystem_path)
+                else:
+                    sync_directory(os.path.dirname(self._filesystem_path))
         else:
             # Delete an item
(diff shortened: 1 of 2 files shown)


liZe:
Looks good for me (at least for posix systems), but I've no experience at all in atomic writes.

@untitaker @kentonv anything to say about this?

radicale/storage.py
```diff
@@ -505,13 +505,38 @@ def update(self, href, vobject_item, etag=None):
...
+                try:
+                    os.rmdir(self._filesystem_path)
+                except OSError:
+                    while True:
+                        tmp_filesystem_path = os.path.join(
+                            os.path.dirname(self._filesystem_path),
+                            ".Radicale.tmp-" + hex(getrandbits(32))[2:])
+                        if not os.path.exists(tmp_filesystem_path):
+                            break
+                    os.rename(self._filesystem_path, tmp_filesystem_path)
```
untitaker: Couldn't we just use the stdlib tmpdir functions for this? I.e. create a tmpdir, move the collection into that folder and delete the tmpdir. Less performant but also less code to overlook.

Also there'd be a race condition here if there was no global lock.

radicale/storage.py
```diff
@@ -505,13 +505,38 @@ def update(self, href, vobject_item, etag=None):
     def delete(self, href=None, etag=None):
         # TODO: use etag in code and test it here
         # TODO: use returned object in code
+        def sync_directory(path):
```
untitaker: I feel like this would be better placed in another module within Radicale.

Maybe I should expose this in atomicwrites?

untitaker:
Unrud's code is always solid, so I only left stylistic nitpicks.

radicale/storage.py
liZe: > Maybe I should expose this in atomicwrites?

It makes sense.

Unrud: Good idea.

Unrud:
> Looks good for me (at least for posix systems),

I can't find any documentation for Windows on this issue. I don't know how to sync meta data or flush the on-disk cache. Replace operations don't seem to be atomic. Moving a file seems to be atomic, as long as the target doesn't already exist and it's within the same volume.
Windows has Transactional NTFS which would be perfect for this, but it's deprecated.

Unrud:
Rebased. I think this can be merged now. If the ``sync_directory`` function gets exposed by ``atomicwrites`` we can later switch to it.

liZe:
Thank you!

PR #454, opened 2016-08-01
Check if file name already exists

Unrud:
The chances are pretty low but maybe the RNG is bad or something

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -25,6 +25,7 @@ entry.
 
 """
 
+import errno
 import json
 import os
 import posixpath
@@ -394,6 +395,14 @@ class Collection(BaseCollection):
 
     @classmethod
     def create_collection(cls, href, collection=None, tag=None):
+        def find_available_file_name(exists_fn):
+            # Prevent infinite loop
+            for _ in range(10000):
+                file_name = hex(getrandbits(32))[2:]
+                if not exists_fn(file_name):
+                    return file_name
+            raise FileExistsError(errno.EEXIST, "No usable file name found")
+
         folder = os.path.expanduser(
             cls.configuration.get("storage", "filesystem_folder"))
         path = path_to_filesystem(folder, href)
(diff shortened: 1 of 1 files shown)


liZe:
I'm not sure about that.

I think that we should:
- put this name generator in a separate function (it's used in many places now),
- be *sure enough* that this function doesn't generate the same name twice.

A *sure enough* solution is generating UUIDs. As it's slow, we prefer using the much faster `getrandbits` function with 2**32 possibilities. We can increase that number just to be 'sure enough'. If the `random` package or the system RNG has a problem, well, that's not our problem :wink:.

- Annual risk of a given person being hit by a meteorite: 6 * 10**-11.
- Risk of having a UUID (122-bit number) collision in a 1000-billion-entries-collection: 9 * 10**-14.
- Risk of having a 32-bit-number collision in a 1000-entries-collection: 1 * 10**-4 (really too high, you're right).

Using 122 instead of 32 gives the same collision rate as UUID, it's OK for me.

Highly reliable benchmark:
- `python -m timeit -c "from random import getrandbits;hex(getrandbits(32))[2:]"`: 1.21 usec per loop,
- `python -m timeit -c "from random import getrandbits;hex(getrandbits(122))[2:]": 1.35 usec per loop,
- `python -m timeit -c "from uuid import uuid4;uuid4().hex": 12.4 usec per loop.

Unrud:
> As it's slow, we prefer using the much faster getrandbits function with 2**32 possibilities.

I don't think that this really matters. It's only used when a calendar is published via webdav and Radicale is written in a slow interpreted language anyway, I would not bother about a few usec in code that is never executed.

Checking if an item with the same name already exists is simple and reliable. Although the PR in it's current state might lead to a endless loop if the RNG is bad and has a very small period.

> Highly reliable benchmark:

On my system the differences are much smaller.

Unrud:
> put this name generator in a separate function (it's used in many places now),

Done.

> Although the PR in it's current state might lead to a endless loop

Fixed.

liZe:
> I don't think that this really matters. It's only used when a calendar is published via webdav and Radicale is written in a slow interpreted language anyway, I would not bother about a few usec in code that is never executed.

Actually, that's the real reason why we did this. Creating big calendars with PUT requests is really slow and using `getrandbits` was a good step ahead.

The idea comes from Kozea/CairoSVG#69 where the difference was even more impressive.

> Checking if an item with the same name already exists is simple and reliable.

OK, let's do this.

Issue #384, opened 2016-04-22
Clarification: How to create a user


untitaker:
Pre-2.0 it was sufficient to request any URL starting with `/<user>/` to create the user. This appears to be no longer possible. Is it advised to create the user manually by creating the folder?

cont. https://github.com/Kozea/Radicale/issues/378

untitaker:
Also, how to create collections?

liZe:
When using `/user/calendar/` paths, we must:
- create the `/user/` WebDAV collection; and then
- create the `/user/calendar/` CalDAV collection (aka calendar).

According to the specification, the first one could be created using a MKCOL request, and the second one with a MKCOL or MKCALENDAR request. The PUT request is also accepted in Radicale and creates collections, even if the WebDAV RFC tells that:

> This specification does not define the behavior of the PUT method for existing collections.  A PUT request to an existing collection MAY be treated as an error (405 Method Not Allowed).
> 
> The MKCOL method is defined to create collections.

Lightning for example sends PUT requests to "publish" calendars. I'm pretty sure that it can't send a MKCOL request (unfortunately, like a lot of other clients).

We can:
- blindly follow the RFC and force users to create their collections with MKCOL requests (and write this in the documentation); or
- create collections and their parents for some requests (including PUT, and well-known GET requests for example) when the user has the right to write; or
- always create collections and their parents (even on GET or REPORT requests) when the user has the right to write.

The third solution is more or less what's in Radicale 1.x and that was both useful and awful (that's why it's been changed).

The first solution is appealing, but I'm afraid of killing the out-of-the-box dark magic of Radicale 1.x. I don't know how many clients will be unable to create collections with this choice.

The second solution is dangerous, but that's the one I prefer.

untitaker:
I don't understand how the second solution significantly differs from the third one.

On 13 July 2016 16:53:03 CEST, Guillaume Ayoub <notifications@github.com> wrote:
>When using `/user/calendar/` paths, we must:
>- create the `/user/` WebDAV collection; and then
>- create the `/user/calendar/` CalDAV collection (aka calendar).
>
>According to the specification, the first one could be created using a
>MKCOL request, and the second one with a MKCOL or MKCALENDAR request.
>The PUT request is also accepted in Radicale and creates collections,
>even if the WebDAV RFC tells that:
>
>> This specification does not define the behavior of the PUT method for
>existing collections.  A PUT request to an existing collection MAY be
>treated as an error (405 Method Not Allowed).
[...]

PR #455, opened 2016-08-01
Creating addressbooks and calendars in DAVdroid

Unrud:
This fixes some issues that prevent creation of collections in DAVdroid from working smoothly.

The first problem is that the properties **addressbook-home-set** and **calendar-home-set** are returned on _/_ with a value of _/_.
The second problem is that the principal collection doesn't exist for new users.

This is also related to #384.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -302,6 +302,23 @@ class Application:
             is_authenticated = self.is_authenticated(user, password)
         is_valid_user = is_authenticated or not user
 
+        # Create principal collection
+        if user and is_authenticated:
+            principal_path = "/%s/" % user
+            if self.authorized(user, self.Collection(principal_path, True),
+                               "w"):
+                with self.Collection.acquire_lock("r"):
+                    principal = next(self.Collection.discover(principal_path),
+                                     None)
+                if not principal or principal.path != principal_path.strip("/"):
+                    with self.Collection.acquire_lock("w"):
+                        # the collection might exist by now
+                        principal = next(self.Collection.discover(
+                            principal_path), None)
+                        if (not principal or
+                                principal.path != principal_path.strip("/")):
+                            self.Collection.create_collection(principal_path)
+
         # Get content
         content_length = int(environ.get("CONTENT_LENGTH") or 0)
         if content_length:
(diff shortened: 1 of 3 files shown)


liZe:
> The first problem is that the properties addressbook-home-set and calendar-home-set are returned on / with a value of /.

Thanks for fixing this, I broke a lot of things rewriting `xmlutils.py`.

> The second problem is that the principal collection doesn't exist for new users.

That's a big problem. I was really surprised yesterday when I realized that iOS Calendar was able to create the user's collection, and I'm not surprised to see that other clients aren't.

Creating the user's collection in `__call__` is probably the best solution to fix #384. @untitaker what do you think of that?

I'm not sure that I understand your changes in `storage.py`. It looks like a workaround to use `discover` to know if the collection already exists. Maybe we can just always call `create_collection` without calling `discover` (and fix `create_collection` to return the collection as soon as we realize that it already exists)?

Unrud:
> I'm not sure that I understand your changes in ``storage.py``. It looks like a workaround to use ``discover`` to know if the collection already exists.

It didn't correctly detect if a collection is a principal or not. This just fixes that, it has nothing to do with using ``discover`` to check if a collection exists.

The problem is cause by ``attributes`` being ``[""]`` for the path ``/``. This only worked at all because ``path_to_filesystem`` ignores empty path parts.

> Maybe we can just always call ``create_collection`` without calling ``discover``

But then we also always need exclusive access to the storage. Checking if the collections exists can be done with a shared lock.

liZe:
> The problem is cause by attributes being [""] for the path /. This only worked at all because path_to_filesystem ignores empty path parts.

OK, I get it. Can't we just add `[0]` and keep the `return`?

> But then we also always need exclusive access to the storage. Checking if the collections exists can be done with a shared lock.

Yes, needing an exclusive lock for each authenticated request is stupid, you're right.

Unrud:
> Can't we just add ``[0]``

Can you clarify this?

> and keep the ``return``?

What's  the intention behind this check? The ``return`` statement was unreachable before.

The semantic of ``discover`` is not changed by this PR, besides that the value of ``principal`` is correct now.

liZe:
> The semantic of discover is not changed by this PR, besides that the value of principal is correct now.

I understand :)

> The return statement was unreachable before.

Of course, I'm stupid. I thought that discovering was useless when `attributes` was empty, but it's not. Sorry for the noise, and thank you for the tests and the pull request!

liZe:
I've simplified some little things in d322d3f, feel free to review!

PR #456, opened 2016-08-02
Properties inside of collection

Unrud:
Move the props file inside of the folder.

  * Creation and deletion of collections can be atomic.
  * The properties file of the root collection is not outside of the
    filesystem_folder.
  * It's easier to delete and move collections by hand.
  * This breaks backward compatibility.

It would be pretty easy to retain backward compatibility by reading props from the old props path if the new file doesn't exist yet and moving the file the next time the props are changed.

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -333,6 +333,8 @@ class Collection(BaseCollection):
         self.path = sanitize_path(path).strip("/")
         self.storage_encoding = self.configuration.get("encoding", "stock")
         self._filesystem_path = path_to_filesystem(folder, self.path)
+        self._props_path = path_to_filesystem(self._filesystem_path,
+                                              "collection.props")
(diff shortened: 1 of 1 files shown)


liZe:
> Creation and deletion of collections can be atomic.

Cool.

> The properties file of the root collection is not outside of the filesystem_folder.

Cool too.

> It's easier to delete and move collections by hand.

Very cool too.

> This breaks backward compatibility.

Who cares :innocent:?

Adding a leading `.` would make the filename not "safe" and we would be able to remove `.endswith(".props")`. I can do that!

Unrud:
>  I can do that!

Sure but then I would name the file something like ``.Radicale.props``. It's consistent with the ``.Radicale.tmp`` files and it's easy to spot what program created them.

liZe:
Done!

PR #458, opened 2016-08-04
Use flock locks for storage locking

Unrud:
These locks are compatible with the command line utility flock,
which comes preinstalled with most Linux distributions.


liZe:
I trust you :smile:!

Issue #442, opened 2016-07-13
Discovery of wrong collection


Unrud:
``Collection.discover(path)`` returns the collection ``path`` if the directory exists. If it doesn't but the parent folder of ``path`` exists, it returns a collection of the parent folder. Otherwise it returns the collection ``path`` ignoring that the folder doesn't exist.

This leads to some problems:

  - Right checking is performed on the  first collection, that ``discover`` returns. If a user calls ``MKCOL /user``, the request is only granted if the user has write access to ``/``. (Supposed ``/user`` doesn't already exist)
  - If the collection ``/user`` exists and the user calls ``GET /user/does_not_exist``, the server correctly replies with 404 but if the collection ``/user`` does not exist, ``discover`` returns an empty dummy collection.

liZe:
Closed by #459.

PR #459, opened 2016-08-04
Improve rights checking and request handlers

Unrud:
* Access rights are checked before the storage is locked and
    collections are loaded.
  * DELETE sends 410 instead of doing nothing or crashing if the target
    doesn't exist.
  * GET always returns 404 if the target doesn't exist.
  * GET doesn't crash if a collection without tag property is requested.
  * MKCOL and MKCALENDAR send 409 if the target already exists.
  * MOVE checks if the target collection of an item actually exists and
    sends 409 otherwise.
  * PUT doesn't crash if a whole collection that doesn't exist yet is
    uploaded and ``content-type`` is ``text/vcard`` or
    ``text/calendar``.
  * PUT distinguishes between simple items and whole collections by the
    following criteria: Target is a collection; Parent exists; Parent
    has the tag property set; Parent contains other items. Before only
    the first two criteria where used, which was very unrelieable. #384
  * PROPPATCH is only allowed on collections and 409 is send otherwise.
  * ``Rights.authorized`` takes a path instead of a collection.
  * ``Collection.discover`` only returns items in ``path``, that
    actually exist. #442

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -29,6 +29,7 @@ should have been included in this package.
 import base64
 import contextlib
 import os
+import posixpath
 import pprint
 import shlex
 import socket
@@ -38,6 +39,7 @@ import subprocess
 import threading
 import wsgiref.simple_server
 import zlib
+from contextlib import contextmanager
 from http import client
 from urllib.parse import unquote, urlparse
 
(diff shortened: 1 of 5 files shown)


liZe:
Amazing. Wow.

I'm resolving the conflict right now.

Unrud:
> I'm resolving the conflict right now.

I can do this, if you want.

liZe:
Fixing all these things is impressive. My favourite is the PUT distinguishing collections from items, I had come to the same conclusion about the criteria, and the PR was already done before I even started to write some code.

Oh, and you even added some tests. And they pass.

Oh, and this pull request doesn't add some lines. It **removes lines**.

I just have to clean a thing or two because of my pathological not-written-yet-but-very-important-style-guide monomania.

Thank you so much.

PR #461, opened 2016-08-05
Remove old code

Unrud:
This was missed while merging #457 and #459.


liZe:
I had missed this one, thanks!

PR #462, opened 2016-08-05
Durable creation of directories and make sure that the root colleciton exists.

Unrud:
This makes sure that all created directories are synced to disk.
Before #457 the root collection was created by ``Collection.acquire_lock`` and before #459 it didn't matter if the folder existed or not.

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -433,7 +433,6 @@ class Collection(BaseCollection):
         if not attributes[0]:
             attributes.pop()
 
-        # Try to guess if the path leads to a collection or an item
         folder = cls._get_collection_root_folder()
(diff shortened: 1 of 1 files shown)


liZe:
Just one thing: `exist_ok` is currently always `True`.

Unrud:
> Just one thing: ``exist_ok`` is currently always ``True``.

I adopted that argument and behaviour from ``os.makedirs``. It can be removed, it always will be ``True``.

liZe:
> I adopted that argument and behaviour from os.makedirs. It can be removed, it always will be True.

OK. I'll remove it and add a line of doc. Thanks!

PR #463, opened 2016-08-06
Atomic MOVE

Unrud:
This makes the MOVE operation atomic and also adds the ability to overwrite items if the **Overwrite** header is set.

MOVE and DELETE are the only remaining operations that are not atomic. This gets fixed by #453 and this PR. All changes besides creation of intermediate collections (which gets fixed by #462) are synced to disk during requests (at least on Linux and MacOS). Calendars and addressbooks should now be pretty safe with Radicale.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -473,8 +473,6 @@ class Application:
         to_path = storage.sanitize_path(to_url.path)
         if not self._access(user, to_path, "w"):
             return NOT_ALLOWED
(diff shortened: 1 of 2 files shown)


liZe:
> This makes the MOVE operation atomic and also adds the ability to overwrite items if the Overwrite header is set.

Perfect.

> MOVE and DELETE are the only remaining operations that are not atomic. This gets fixed by #453 and this PR.

Thank you, thank you, and thank you again! :tada: :cake: :candy:.

PR #464, opened 2016-08-07
Some clients expect collections to end with /

Unrud:
Compatibility with InfCloud/CalDavZAP/CardDavMATE


Unrud:
Fix is included in #466

Issue #29, opened 2013-05-16
Add tests


liZe:
Radicale should be tested. There's only one test sending a GET request at `/`, and that's a shame.

DAViCal has for example [an excellent testing suite](http://repo.or.cz/w/davical.git/tree/HEAD:/testing/tests/regression-suite).

untitaker:
I don't think using litmus is a realistic option ATM, primarily because it asserts full compliance to the RFCs. @liZe If you want I'll whip up a PR which downloads my CalDAV/CardDAV client and runs its (very basic) testsuite against Radicale. That's how I found all the bugs I've reported in the past.

untitaker:
Bump. I'd be willing to prepare a PR that installs vdirsyncer into Travis, and runs its testsuite against Radicale. The diagnostics and debugging are ugly at the moment, but it does find bugs!

untitaker:
@liZe While I have you on the line, could you respond to my comments above?

liZe:
It's OK for me!

liZe:
Tests seem to be quite good for now, I think that it's enough for 2.0.0.

PR #466, opened 2016-08-07
Set correct path for child collections

Unrud:
This also adds a test for PROPFIND. In general we need more tests to catch stupid mistakes like this.


liZe:
> In general we need more tests to catch stupid mistakes like this.

Yes, #29 is not closed yet :wink:.

PR #467, opened 2016-08-08
Create predefined calendars and addressbooks for new users

Unrud:
The administrator can define a list of collections (calendars and addressbooks) in the config file that are created for new users. This is especially useful for clients that don't support the creation of collections.

### DIFF
--- a/config
+++ b/config
@@ -113,6 +113,13 @@
 # Example: git add -A && (git diff --cached --quiet || git commit -m "Changes by "%(user)s)
 
 
+[skel]
+
+# Collections that are created for new users separated by a comma
+#addressbooks = addressbook.vcf
+#calendars = calendar.ics
+
+
 [logging]
 
 # Logging configuration file
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -287,8 +287,7 @@ class Application:
             if self.authorized(user, principal_path, "w"):
                 with self._lock_collection("r", user):
                     principal = next(
-                        self.Collection.discover(principal_path, depth="1"),
-                        None)
+                        self.Collection.discover(principal_path), None)
                 if not principal:
                     with self._lock_collection("w", user):
                         self.Collection.create_collection(principal_path)
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -290,7 +290,7 @@ class Application:
                         self.Collection.discover(principal_path), None)
                 if not principal:
                     with self._lock_collection("w", user):
-                        self.Collection.create_collection(principal_path)
+                        self.Collection.create_user(user)
 
         # Verify content length
         content_length = int(environ.get("CONTENT_LENGTH") or 0)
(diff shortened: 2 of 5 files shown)


liZe:
I'm not the biggest fan of this pull request. All the clients I've tested (even Thunderbird) are able to create calendars and addressbooks if the user's root collection has been created. Do you know a client that it helps?

Unrud:
I wouldn't say that Thunderbird(/Outlook) supports creation of collections.

  - Create a local calendar
  - Publish that calendar via WebDAV
  - Delete it locally
  - Add the calendar via CalDAV

That's more like a hack or is there another method?
With this PR you can just add http://server/**username**/calendar.ics and you are done.

Clients that support creation like DAVdroid also have there flaws. The created calendars have a nice displaynames but the collection is a UUID and it doesn't even show the real URL. If you want to add that calendar in a client that doesn't support discovery (like Thunderbird or Outlook) you have to lookup the directory name on the server. If you don't have access to the server, you have to do a PROPFIND request by hand via ``curl`` or something similar. Am I missing something here?

There are also clients that don't support creation of collection at all (like InfCloud).

This PR only solves part of this problem, it's still difficult the create and use additional collections.
A web interface to show, create and delete collections would be nice.

rfc2822:
> Clients that support creation like DAVdroid also have their flaws. The created calendars have a nice displayname but the collection is a UUID

That design decision was made because
* users don't know the details of what is allowed for collection names, and will often use special characters, which then often cause problems,
* they shouldn't have to bother with collection names, because [that's the purpose of `displayname`](https://tools.ietf.org/html/rfc4791#section-5.3.1): "Clients SHOULD use the DAV:displayname property for a human-readable name of the calendar.".

> and it doesn't even show the real URL.

At the moment, not directly (we have thought about a per-collection info popup). However, as a workaround, you can go to DAVdroid / App Settings / Debug info and you will see all collections with URL and name in the database dump.

Unrud:
> That design decision was made because

I did not question any design decisions. The problem is that various clients (and servers) support the protocol to various degrees. In Thunderbird/Outlook you have to enter the URL of the collection directly, that's of course not the fault of DAVdroid.

> you can go to DAVdroid / App Settings / Debug info and you will see all collections with URL and name in the database dump.

OK, but that's not very user friendly and you still have to get the URL from your phone to your computer (where Thunderbird/Outlook is installed). Radicale should have a web interface to show collections with names and URLs.

rfc2822:
@Unrud No problem, I just wanted to mention the Debug info possibility. You can "share" the debug  info by the way, so you can send it to your PC over email, for instance.

liZe:
> That's more like a hack or is there another method?

There's no other method, and that's awful for sure. But automatic creation of collections is probably a bad idea too (see #384 for example).

> A web interface to show, create and delete collections would be nice.

I'd really prefer that solution!

Unrud:
Closed in favour of #484.

PR #471, opened 2016-08-10
Fix logger

Unrud:
9192a77 fixes a problem that can cause a crash, depending on the content of ``answer``.
The rest is just cosmetics.


liZe:
Thank you!

PR #477, opened 2016-08-11
Remove etags from raw data

Unrud:
Remove all etags that are directly calculated from data that's read from files.
  * They are not used anywhere (luckily).
  * Etags that are send to clients are calculated from the output of vobject's serialize method. If files are edited externally and vobject normalizes them (like wrapping long lines or replacing all line endings by \r\n), the etags that are sent to the client and the etags that are calculated from raw data will never match. If a new version of vobject is released and the formatting changes slightly, the checks will also always fail.

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -306,7 +306,7 @@ class BaseCollection:
         the filters and this implementation.
         This returns all event by default
         """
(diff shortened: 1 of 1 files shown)


liZe:
Thanks! It had to be cleaned, these parameters come ancient times :smile:.

PR #478, opened 2016-08-11
Correctly write line endings to files

Unrud:
vobject uses ``\r\n`` as line endings. Writing this to a file is not a problem on Linux and newer versions of MacOS. On Windows ``\r\r\n`` gets written to disk and on older versions of MacOS ``\r\r`` gets written to disk, because python replaces \n by the system depended line ending.

Together with #476 all tests are working on Windows.


liZe:
> Together with #476 all tests are working on Windows.

:tada:! :lollipop:! :gift:!

PR #480, opened 2016-08-12
Ask for authentication if current-user-principal is requested

Unrud:
DAVdroid dropped support for preemptive authentication in version 1.2.3. Currently it's impossible to login as a user.
Returning the DAV:unauthenticated pseudo-principal as specified in RFC 5397 doesn't work for DAVdroid (or maybe I was just doing something wrong).

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -532,9 +532,12 @@ class Application:
             items = itertools.chain([item], items)
             read_items, write_items = self.collect_allowed_items(items, user)
             headers = {"DAV": DAV_HEADERS, "Content-Type": "text/xml"}
-            answer = xmlutils.propfind(
+            status, answer = xmlutils.propfind(
                 path, content, read_items, write_items, user)
-            return client.MULTI_STATUS, headers, answer
+            if status == client.FORBIDDEN:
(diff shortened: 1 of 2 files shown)


rfc2822:
> DAVdroid dropped support for preemptive authentication in version 1.2.3. Currently it's impossible to login as a user.

This is not quite true. [DAVdroid now sends Basic authentication preemptively by default, but only for HTTPS connections.](https://gitlab.com/bitfireAT/dav4android/blob/ff40bfd61f40a445b0e9414ae15b4a02fb95f64b/src/main/java/at/bitfire/dav4android/BasicDigestAuthHandler.java#L71) The user setting for preemptive authentication is not needed anymore, because in any case, 401 responses are interpreted correctly (previous behaviour: 401 was only interpreted correctly if preemptive auth. was disabled). This means:

* HTTPS: Basic digest auth is preemptively sent with first request. When 401 is received, supported authentication mechanisms (namely Digest) are tried, if possible.
* Plain HTTP: first request is sent without preemptive Basic auth, because it's unacceptable to transfer unencrypted passwords preemptively. When 401 is received, supported authentication mechanisms (namely Digest) are tried, if possible.

PR #484, opened 2016-08-18
Web interface

Unrud:
It's possible to view, edit and create calendars and addressbooks in the principal folder. The web interface also includes [InfCloud](https://www.inf-it.com/open-source/clients/infcloud/) for managing events, tasks and contacts.

The web interface lives in it's own module. This PR only changes **do_GET** to forward requests on _/.web_ to the ``RadicaleWeb`` module and to redirect _/_ to _/.web_, if the module is available. (Otherwise the normal ``Radicale works!`` message is shown.)

The web interface module can be found here: https://github.com/Unrud/RadicaleWeb

The design could be improved, I just copied the style sheets from http://radicale.org for now.

This renders #467 obsolete.


liZe:
Having a web interface is a good idea, but I'd really like to have it in a separate package!

Unrud:
> but I'd really like to have it in a separate package!

Sure, what did you have in mind? Do you want to ship it with Radicale or should users install it separately?

liZe:
> Do you want to ship it with Radicale or should users install it separately?

Separately is probably the best solution. I'm sure that a web interface could be really powerful and really useful, but I don't feel like I'm able/motivated to maintain a project like this, Radicale is big enough :wink:.

liZe:
@Unrud Do you have plans to distribute the web interface separately?

Unrud:
I've moved the web interface into it's own module. This PR now only contains the hook to integrate the module into Radicale.

The web interface module can be found here: https://github.com/Unrud/RadicaleWeb

> I'd love to see the scope of the web ui broadened to actual card/event editing/management.

I've added [InfCloud](https://www.inf-it.com/open-source/clients/infcloud/) to the web interface.

Unrud:
> I've added [InfCloud](https://www.inf-it.com/open-source/clients/infcloud/) to the web interface.

InfCloud doesn't work properly without #565. It can't detect changes of collections without sync-token.

Unrud:
Fixed the case that SCRIPT_NAME is set to something like ``/sub/path`` and the requested URL is ``/sub/path``. The browser was redirected to ``/sub/.web`` instead of ``/sub/path/.web``.
(Part of the fix is the same as #598)

PR #489, opened 2016-08-25
Try to decode URLs with utf-8 (Fixes #486)

### LINKED ISSUE #486: URL encoding
According to [RFC 3986](https://tools.ietf.org/html/rfc3986) (2005) URLs should be encoded with **utf-8**.
The integrated server uses ``wsgiref.WSGIRequestHandler`` which decodes URLs using **latin1**.
At least DAVdroid and Thunderbird use **utf-8** and all URLs containing non-ascii characters don't work.

Unrud:


liZe:
Cool, thank you!

PR #490, opened 2016-08-25
Move hook into storage.Collection

Unrud:
The hook is only valid for filesystem storage, it's meaningless for other backends like databases.


liZe:
Well, that's finally not a bad idea. If some backends want to use this feature too, they can, but they can also ignore it if they want/need too.

PR #491, opened 2016-08-25
Optimize upload of whole collections

Unrud:
This gets rid of the useless atomic creation of items inside a new temporary collection. Additionally all calls to fsync are bundled and the directory is only synced once.

Result from a simple speed test (addressbook with 1000 entries):
  * Before: 1:25 minutes
  * After: 0:05 minutes

I moved ``upload_all`` from ``BaseCollection`` to ``Collection`` because it's not used anywhere else and probably never will be as WebDAV doesn't support bulk uploads. Additionally I renamed it to ``upload_all_nonatomic`` to reflect it's new usage.

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -552,7 +552,7 @@ class Collection(BaseCollection):
                         items.extend(
                             getattr(collection, "%s_list" % content, []))
                     items_by_uid = groupby(sorted(items, key=get_uid), get_uid)
-                    collections = {}
+                    vobject_items = {}
                     for uid, items in items_by_uid:
                         new_collection = vobject.iCalendar()
                         for item in items:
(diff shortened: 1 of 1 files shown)


liZe:
> This gets rid of the useless atomic creation of items inside a new temporary collection. Additionally all calls to fsync are bundled and the directory is only synced once.

Really good idea! Thanks!

> I moved upload_all from BaseCollection to Collection because it's not used anywhere else and probably never will be as WebDAV doesn't support bulk uploads.

I'm not that sure. Having an `upload_all` function for a database storage is for example an easy way to put all the interts into one transaction.

> Additionally I renamed it to upload_all_nonatomic to reflect it's new usage.

I think that it's an implementation detail, not something that should be exposed in the API.

Unrud:
> I'm not that sure. Having an ``upload_all`` function for a database storage is for example an easy way to put all the interts into one transaction.

``upload_all`` cannot be implemented atomic in the filesystem backend, without copying the whole collection. I renamed it to ``upload_all_nonatomic`` to make clear that it's not atomic.

I can add an atomic implementation of ``upload_all`` which uses ``upload_all_nonatomic`` internally if you want. But *DAV can't change multiple resources with one request and it will not be used. I suggest that we add it when we find a use case for it

> I think that it's an implementation detail, not something that should be exposed in the API.

It's not just an implementation detail. ``upload_all_nonatomic`` must not be used outside of temporary (new) collections, because it's not atomic.

liZe:
I'm convinced, thank you!

PR #492, opened 2016-08-25
Remove BaseCollection.update

Unrud:
I don't think that this can be used for optimizations.

It's useless in the filesystem backend, **SQL** has ``REPLACE`` and I doubt that there is much use in any other storage mechanism.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -600,10 +600,7 @@ class Application:
                 if tag:
                     parent_item.set_meta({"tag": tag})
                 href = posixpath.basename(path.strip("/"))
-                if item:
-                    new_item = parent_item.update(href, items[0])
-                else:
-                    new_item = parent_item.upload(href, items[0])
+                new_item = parent_item.upload(href, items[0])
             headers = {"ETag": new_item.etag}
             return client.CREATED, headers, None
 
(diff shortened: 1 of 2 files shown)


liZe:
> It's useless in the filesystem backend, SQL has REPLACE and I doubt that there is much use in any other storage mechanism.

By `REPLACE` you mean [`MERGE`](https://en.wikipedia.org/wiki/Merge_(SQL)), don't you? `MERGE` doesn't have the same syntax in the different SQL databases, I think that it's a good reason to keep the method.

Unrud:
> By ``REPLACE`` you mean ``MERGE``, don't you?

I actually meant [``REPLACE``](https://dev.mysql.com/doc/refman/5.7/en/replace.html) but it's only available in MySQL and SQLite (which [TimeRangeIndexedStorage](https://github.com/Kozea/TimeRangeIndexedStorage) uses).

> ``MERGE`` doesn't have the same syntax in the different SQL databases

That's annoying but for a general purpose SQL backend it's probably a good idea to use an abstraction layer like SQLAlchemy which supports [merging](http://docs.sqlalchemy.org/en/latest/orm/session_state_management.html#merging).

liZe:
I'm not sure about that, and it's probably the first time I'm not really happy about removing lines from Radicale! I'll merge the pull request, we'll see if anyone really needs this method.

PR #496, opened 2016-08-30
Remove empty line in log

Unrud:
Separating requests by an empty line doesn't work any longer with parallel (overlapping) requests.


liZe:
True!

PR #499, opened 2016-08-30
More secure external authentication

Unrud:
It's unnecessary to check users from external authentication with ``is_authenticated``.
The ``auth`` module must accept users with empty passwords to make this work. That's a bit dangerous since anyone can login with a empty password unless the web server filters the **AUTHORIZATION** header.

### DIFF
--- a/config
+++ b/config
@@ -83,6 +83,10 @@
 # Value: plain | sha1 | ssha | crypt | bcrypt | md5
 #htpasswd_encryption = crypt
 
+# Get users from the REMOTE_USER environment variable instead of
+  using HTTP authentication
+#remote_user = False
+
 
 [rights]
 
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -336,9 +336,10 @@ class Application:
             login, password = self.decode(base64.b64decode(
                 authorization.encode("ascii")), environ).split(":", 1)
             user = self.Auth.map_login_to_user(login)
+            is_authenticated = self.Auth.is_authenticated(user, password)
         else:
             user = self.Auth.map_login_to_user(environ.get("REMOTE_USER", ""))
-            password = None
+            is_authenticated = bool(user)
 
         # If "/.well-known" is not available, clients query "/"
         if path == "/.well-known" or path.startswith("/.well-known/"):
@@ -348,8 +349,6 @@ class Application:
             # Prevent usernames like "user/calendar.ics"
             self.logger.info("Refused unsafe username: %s", user)
             is_authenticated = False
-        else:
-            is_authenticated = self.Auth.is_authenticated(user, password)
         is_valid_user = is_authenticated or not user
 
         # Create principal collection
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -331,15 +331,14 @@ class Application:
 
         # Ask authentication backend to check rights
         authorization = environ.get("HTTP_AUTHORIZATION", None)
-        if authorization and authorization.startswith("Basic"):
+        login = password = ""
+        if self.configuration.getboolean("auth", "remote_user"):
+            login = environ.get("REMOTE_USER", "")
+        elif authorization and authorization.startswith("Basic"):
             authorization = authorization[len("Basic"):].strip()
             login, password = self.decode(base64.b64decode(
                 authorization.encode("ascii")), environ).split(":", 1)
-            user = self.Auth.map_login_to_user(login)
-            is_authenticated = self.Auth.is_authenticated(user, password)
-        else:
-            user = self.Auth.map_login_to_user(environ.get("REMOTE_USER", ""))
-            is_authenticated = bool(user)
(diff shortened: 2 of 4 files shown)


jaesivsm:
The previous approach to this was through an auth type. This way, the administrator could decide whether or not to trust "REMOTE_USER". 
I did put together an attempt at a [plugin](https://github.com/jaesivsm/radicale_auth_remote_user) for radicale that would provide the same result while still let the same kind of choice than the "auth type" did in Radicale 1.x

If this method is preferred, I'll drop my module.

liZe:
> The auth module must accept users with empty passwords to make this work.

As far as I can remember, it was a feature. But it was a looooong time ago.

> That's a bit dangerous since anyone can login with a empty password unless the web server filters the AUTHORIZATION header.

Allowing users with no password can be a useful feature for some authentication modules like htpasswd. Now that we allow external authentication modules, I think that it could be even more useful.

Unrud:
> The previous approach to this was through an auth type.

Interesting. The ``remote_user`` module is basically the same as ``None``. It grants access to everyone (``return user is not None``).

The auth module doesn't know if ``user`` comes from the **AUTHORIZATION** header or from the **REMOTE_USER** environment variable. You could do something like ``return password is None``, but it exploits an implementation detail.

> This way, the administrator could decide whether or not to trust "REMOTE_USER". 

If this is a concern it would be better to add a configuration option.

> I did put together an attempt at a [plugin](https://github.com/jaesivsm/radicale_auth_remote_user) for radicale that would provide the same result while still let the same kind of choice than the "auth type" did in Radicale 1.x

This has the same problem as the  module from Radicale 1.x.

> Allowing users with no password can be a useful feature for some authentication modules like htpasswd.

But this is only about users that are already authenticated by the web server. Currently ``password`` is set to ``None`` for remote users. Depending on the encryption method the htpasswd auth module will either always return ``False`` or throw an exception.

jaesivsm:
> The remote_user module is basically the same as None. It grants access to everyone (return user is not None).

Well, not really. `NoneAuth` won't care if you provide a username or not. Anonymous request will be considered as authenticated for example. Of course you could argue that unauthenticated request will be blocked ahead and won't be passed to Radicale so it won't matter. It'll certainly be true in most case, but I can imagine a configuration where, for some reason, anonymous request are passed on to Radicale which has to handle them.

> The auth module doesn't know if user comes from the AUTHORIZATION header or from the REMOTE_USER environment variable.

Yeah, I agree with that. Obtaining the username could also be relevant of the auth module, so it could be overridden by the plugin.

> If this is a concern it would be better to add a configuration option.

I have no problem with that solution.

I think setting remote_user as an auth method is a more explicit way to say that you don't want radicale to handle authentication and want it to trust user obtained through the request context.

Unrud:
> Well, not really. ``NoneAuth`` won't care if you provide a username or not. Anonymous request will be considered as authenticated for example.

It doesn't matter what the auth plugin returns for anonymous users.

Unrud:
> If this is a concern it would be better to add a configuration option.

I added the configuration option ``remote_user`` which lets the user decide if the web server or Radicale should handle authentication.

> Allowing users with no password can be a useful feature for some authentication modules like htpasswd.

I've restored the original state, but set the password to an empty string instead of ``None`` to prevent the exceptions.

jaesivsm:
> It doesn't matter what the auth plugin returns for anonymous users.

Yeah I realized that, NoneAuth actually does some authentication since reads from the `HTTP_AUTHENTICATION`.
Wouldn't it be more logic if `NoneAuth` didn't actually do any auth and just retrieve a username from the `PATH` ?

Unrud:
> Yeah I realized that, ``NoneAuth`` actually does some authentication since it reads from the ``HTTP_AUTHENTICATION`` environ and extract a user out of it.

No, it doesn't? The auth modules just check user and password.

> Wouldn't it be more logic if ``NoneAuth`` didn't actually do any auth and just retrieved a user from the ``PATH_INFO`` ?

That's a bad idea because principal collections are automatically created.

jaesivsm:
> > Yeah I realized that, NoneAuth actually does some authentication since it reads from the HTTP_AUTHENTICATION environ and extract a user out of it.

> No, it doesn't? The auth modules just check user and password.

Sorry I said it wrong. I meant that when you select `NoneAuth` credentials are still read and a logic based on users is still applied. One could assume that `NoneAuth` would allow you to throw request at Radicale without any auth header in them.

> That's a bad idea because principal collections are automatically created.

Ok then.

Unrud:
This is only a problem if client side certificates are used ([like here](https://forums.bitfire.at/topic/380/support-for-ssl-client-certificates/15)) or the WSGI upstream server is directly reachable (because of misconfiguration).

Nginx and Apache filter the **AUTHORIZATION** header by default, most other web servers with WSGI support don't seem to support client side certificates anyway.

This PR only improves security slightly in case of misconfiguration, I will close it because it's not worth it.

PR #500, opened 2016-08-30
Map logins to internal users in Auth module

Unrud:
This makes it possible to implement #349 as a ``Auth`` module. Another use case would be to encode usernames that contain characters unsupported by the file system.


liZe:
Really simple and really useful, thanks!

PR #501, opened 2016-08-30
HTTP error messages

Unrud:
Browsers just show a blank page if an error occurs. You have to open the developer tools to see the HTTP status code. E.g. a user wants to download a calendar in the browser and the URL is wrong.
Some tools like curl don't show any indication of an error.

Some additional improvements concerning HTTP errors:

  * Return a error message instead of a blank page for GET requests on directories .(The same behaviour as most web servers.)
  * Don't use the GONE status code as Thunderbird doesn't recognize it correctly and shows synchronization errors.


liZe:
Radicale looks like a real server now, even for browsers :smile:.

PR #502, opened 2016-08-31
Plain text for GET / instead of malformed HTML

Unrud:
It's basically the same in browsers and looks nicer in tools that don't support HTML rendering (like curl).


liZe:
Why not.

Just for the record: it's not malformed at all, it's valid HTML5 :wink:.

PR #504, opened 2016-09-01
obtaining creds from env can now be overridden in plugins

jaesivsm:
Following the discussion in #499 I extracted the credentials obtaining from the main module to the auth module so that it can be overridden in a plugin.

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -208,30 +208,6 @@ class Application:
 
         return request_environ
 
-    def decode(self, text, environ):
-        """Try to magically decode ``text`` according to given ``environ``."""
-        # List of charsets to try
-        charsets = []
-
-        # First append content charset given in the request
-        content_type = environ.get("CONTENT_TYPE")
-        if content_type and "charset=" in content_type:
-            charsets.append(
-                content_type.split("charset=")[1].split(";")[0].strip())
-        # Then append default Radicale charset
-        charsets.append(self.encoding)
-        # Then append various fallbacks
-        charsets.append("utf-8")
-        charsets.append("iso8859-1")
-
-        # Try to decode
-        for charset in charsets:
-            try:
-                return text.decode(charset)
-            except UnicodeDecodeError:
-                pass
-        raise UnicodeDecodeError
-
     def collect_allowed_items(self, items, user):
         """Get items from request that user is allowed to access."""
         read_allowed_items = []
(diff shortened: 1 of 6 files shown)


Unrud:
I'm not sure if this is really useful. Either Radicale handles authentication by itself or leaves it to the web server (REMOTE_USER env. var.), there are no other options that could be implemented in plugins.

jaesivsm:
Well it allows to clearly define where you take your user from. I don't think they are use cases where both `HTTP_AUTHENTICATION` and `REMOTE_USER` are used at the same. Such case would not be actually supported indeed.

As this feature was removed when storages, auths and other stuffs like that were cleaned I assumed Radicale was wanted to work in a single way. That's why plugins were made available.

> there are no other options that could be implemented in plugins.

I can imagine other auth mechanism overriding that part also. Leaving the possibility open doesn't seem harmful to me. 

But again, I do not stand by that solution firmly.

Unrud:
> I don't think they are use cases where both HTTP_AUTHENTICATION and REMOTE_USER are used at the same.

True. I think that a configuration option to switch between both would be enough.

> I can imagine other auth mechanism overriding that part also.

Do you have an example?

jaesivsm:
> Do you have an example?

Well not really. But for one the http auth mechanism doesn't have to be `Basic` (see https://tools.ietf.org/html/rfc2617#section-3) and I can imagine user implementing kerberos token over that.

Unrud:
> But for one the http auth mechanism doesn't have to be ``Basic`` (see https://tools.ietf.org/html/rfc2617#section-3)

If you want implement anything besides ``Basic``, you also have to expose the response headers to the module.

> I can imagine user implementing kerberos token over that.

Sure, you can also just use a web server that handles this.

jaesivsm:
> If you want implement anything besides Basic, you also have to expose the response headers to the module.

https://github.com/Kozea/Radicale/pull/504/commits/568f804da7781492e60279c49d925b7a14a8f50a#diff-058c90db22e1ad548ea5699e233221b3R92

> Sure, you can also just use a web server that handles this.

As for Basic auth, or parsing htpasswd files or many other stuffs that are implemented in Radicale. I don't see your point.

Unrud:
> As for Basic auth, or parsing htpasswd files or many other stuffs that are implemented in Radicale. I don't see your point.

If you want to use something complicated like Kerberos it's much easier to use a web server that already supports it. The same is true for other less-used authentication methods like client-side certificates.

I think that Basic and Digest are the only widely supported methods. Digest could be implemented in addition to Basic (without a plug-in module), on the other hand you can also just enable SSL which is much more secure.

jaesivsm:
This way we would have Basic, Digest, and the remote user half implemented in the default auth module (since the other part is the user retrieving from the context and it would take place in the main code). That looks a bit v1-ish to me.

I thought the whole point of splitting into main project and plugins was to avoid this. It's a design choice.

Unrud:
My concern is that this PR adds complexity without any real gain. In it's current state it's not even possible to extent Radicale with a plug-in  module beyond Basic auth and REMOTE_USER. Client side certificates are even impossible with this approach, you would need to extent the integrated server.

I have the feeling that nobody will ever use it, and instead just setup **nginx** or similar.

jaesivsm:
> My concern is that this PR adds complexity without any real gain.

I'm moving code around, removing auth retrieving from the main module and pasting it into the auth module. I didn't really write any new code so no complexity is added; the code is just more organized. 

If you don't think that part of the code belong there and you want more control structure in the main module to deal with `REMOTE_USER`s and configuration then I don't think we can come to an agreement.

But again that's a very minor change anyway.

Unrud:
Implemented in #612

PR #507, opened 2016-09-02
 Repair and test authentication

Unrud:


radicale/tests/test_base.py
```diff
@@ -766,6 +766,14 @@ def test_authorization(self):
         assert status == 207
         assert "href>/user/<" in answer
 
+    def test_authentication(self):
+        """Test if server sends authentication request."""
+        self.configuration.set("rights", "type", "owner_only")
+        self.application = Application(self.configuration, self.logger)
```
liZe: Changing `self.application` in the test is a bit odd, maybe we should put this test elsewhere.

Unrud: I'm not sure. ``self.configuration`` is also changed, why draw the line at ``self.application``? Putting the test somewhere else would involve duplicating the whole fixture, just to move one line from the ``setup`` method into the test method.

liZe: Hmmm. Let's keep it here for now.

PR #510, opened 2016-09-03
Improve formatting of log messages

Unrud:
Adds thread and level in simple log formatter and to the default formatter. The console log is not very useful without this additional information, when requests overlap.

Slightly changes the format:
Old: ... - 139891659843328 - INFO: Starting Radicale
New: ... - [7f3b10b12700] INFO: Starting Radicale

### DIFF
--- a/logging
+++ b/logging
@@ -53,7 +53,7 @@ formatter = full
 
 [formatter_simple]
 # Simple output format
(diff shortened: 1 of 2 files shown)


liZe:
Nice!

PR #512, opened 2016-09-04
Cleanup/Improve storage.py

Unrud:
a12ef69 and  5dbf9df add missing checks for safe conversion of paths to the local filesystem. They are only relevant on Windows and currently it's not possible to exploit it.

 6df54bf adds a log message that includes the name of a faulty component. Before only the exception from vobject showed up in the log but the filename of the component that caused the error was missing.

The Collection.get method stripped {} from `href`, 03fbb1e removes that. If someone uploaded a component that starts or ends with { or }, all REPORT requests on that collection failed and it was impossible to delete the component.
@liZe Why was that there in the first place? I couldn't find something like this in the relevant RFCs and hope removing it doesn't break some client.

de09f66 changes last_modified to only look at relevant files.

All other changes are just cosmetic.

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -142,7 +142,8 @@ def is_safe_path_component(path):
 
 
 def is_safe_filesystem_path_component(path):
-    """Check if path is a single component of a filesystem path.
+    """Check if path is a single component of a local and posix filesystem
+       path.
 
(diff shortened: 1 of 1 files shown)


liZe:
Thanks a lot!

PR #513, opened 2016-09-04
Remove base_prefix

Unrud:
This removes the `base_prefix` and `can_skip_base_prefix` configuration options. Instead the `SCRIPT_NAME` environment variable is used according to PEP 333.

For reverse proxies the HTTP header `X-Script-Name` is introduced which allows overwriting of the `SCRIPT_NAME` variable ([Source](http://flask.pocoo.org/snippets/35/)). This solution is cleaner and also allows multiple reverse proxies simultaneously.

Example Nginx configuration:

```
location /sub/folder/radicale {
    proxy_pass localhost:5232/; # The / is important!
    proxy_set_header X-Script-Name /sub/folder/radicale;
}
```

This fixes some additional smaller issues with URL handling:
- The request URL was double unquoted: `%2525` evaluated to `%` instead of `%25`.
- URLs from XML requests were not sanitized
- Invalid URLs in XML requests were just silently ignored. This PR adds some log messages.
- `name_from_path` function didn't check that the name is valid. For example it could return something like `user/calendar.ics`.
- hreferences in REPORT responses were not expanded with `base_prefix`
- hreferences in XML responses were not quoted. [RFC 4918](https://tools.ietf.org/html/rfc4918) states that they are URIs and [RFC 3986](https://tools.ietf.org/html/rfc3986) says that URIs must always be in percent-encoded form.

### DIFF
--- a/config
+++ b/config
@@ -51,12 +51,6 @@
 # Reverse DNS to resolve client address in logs
 #dns_lookup = True
 
-# Root URL of Radicale (starting and ending with a slash)
-#base_prefix = /
-
-# Possibility to allow URLs cleaned by a HTTP server, without the base_prefix
-#can_skip_base_prefix = False
-
 # Message displayed in the client when a password is needed
 #realm = Radicale - Password Required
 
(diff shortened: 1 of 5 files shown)


liZe:
That's perfect, thanks a lot.

PR #517, opened 2016-09-17
try/catch of objects avoiding not serving any object in case an object has no valid contents

pbiering:
detected the need of try/catch while having a broken ICS file

also log duration of how many objects are read per folder for statistical/diagnostic purposes


pbiering:

PR #518, opened 2016-09-17
improved request logging

pbiering:
log client IP address (X-Forwarded-For has precedence), User Agent and duration in consistent log level (if given, also the HTTP_DEPTH value)


pbiering:

PR #519, opened 2016-09-17
filter broken vcard entries before delivered, found on real life syst…

pbiering:
…em at least 2 cards where vobject claimed about

 vobject.base.ValidateError: 'VCARD components must contain at least 1 N'

fix is a workaround, probably there is a better solution.

Problem is that vcard has 2 different ways to be read, one is via storage.py, the other is for propfind directly via xmlutils.py

Also my two (anoymized) real-life exampes stored in tests/static


pbiering:

PR #536, opened 2016-11-13
fix encoding

rabbitcdn:
SyntaxError: Non-ASCII character '\xc2' in file.....


liZe:
Oh… The current version of Radicale is only compatible with Python 3.

PR #540, opened 2016-11-30
remove umask(0) in daemon mode

twoodcock:
Can we please remove the umask(0) in daemon mode?

I don't like having files left with world write access. World read is unnecessary for this daemon as well.

Without a configuration option that allows an administrator to specify the umask, I think simply removing this umask makes sense.

Resetting the umask to zero might make sense in other cases. In this case, it is not necessary. In fact, it stops an administrator from properly implementing permissions.


pbiering:
sounds reasonable, merged into my pull request now

liZe:
Thank you!

PR #542, opened 2016-12-29
Add auth based on pamela (https://github.com/minrk/pamela).

YtvwlD:
Most PAM libraries for Python are outdated.
Most of them don't even work with Python 3.
Pamela does.

The auth module is mostly based on the existing one for PAM
but is slightly changed to be used with pamela.

### DIFF
--- a/config
+++ b/config
@@ -73,7 +73,7 @@
 [auth]
 
 # Authentication method
-# Value: None | htpasswd | IMAP | LDAP | PAM | courier | http | remote_user | custom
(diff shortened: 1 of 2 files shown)


YtvwlD:
Does it make sense to declare [pamela](https://github.com/minrk/pamela) as a dependency? It is only used when this auth method is enabled.

Unrud:
All authentication modules besides `htpasswd` were removed in Radicale 2.0.0.
Other methods can be implemented as a [plugins](http://radicale.org/plugins/).

PR #558, opened 2017-03-07
Check for conflicting file names

Unrud:
On Windows file systems the user "TESTUS~1" can access the data of the user "testuser".


liZe:
> On Windows file systems the user "TESTUS~1" can access the data of the user "testuser".

Wow.

PR #559, opened 2017-03-07
Cleanup remote host logging

Unrud:
* WSGIRequestHandler doesn't set REMOTE_HOST if dns lookup is disabled.
* Also log the forwarding host for forwarded requests


liZe:
That's much better, thank you.

PR #560, opened 2017-03-07
Always close files when creating collection

Unrud:
If an exception occurs we rely on garbage collection to close the files.


liZe:
I trust you for this one :wink:.

PR #564, opened 2017-03-10
Use logger for WSGIServer

Unrud:
Log exception from the WSGIServer. Exceptions from socket timeouts are currently written to stderr.


liZe:
:heart:.

PR #565, opened 2017-03-10
Support for sync-token and sync-collection

### LINKED ISSUE #306: Support for sync-collection REPORT
After discussing with the author of [CalDavZAP](https://www.inf-it.com/open-source/clients/caldavzap/), [CardDavMATE](https://www.inf-it.com/open-source/clients/carddavmate/) and [InfCloud](https://www.inf-it.com/open-source/clients/infcloud/), I think that Radicale lacks an important feature : support for sync-collection REPORT (with sync-tokens).

This is how InfCloud react when a server does not support the **sync-collection REPORT** feature :
- do
  - download a list of _ALL_ URLs+Etags for all objects **=> bandwidth consuming**
  - if there is an URL not already downloaded then this is a new object, then download it **=> cpu consuming**
  - if there is an URL which is already downloaded but the Etag is changed, this object was updated, then re-download it  **=> cpu consuming**
  - if there is an URL which is already downloaded but no more present on the server, this object was deleted, then delete it **=> cpu consuming**
- while(true);

This is how InfCloud react when a server **does** support the **sync-collection REPORT** feature (such as DAViCal) :
- download a list of _ALL_ URLs+Etags + get the sync-token **=> bandwidth consuming** but **only once**
- do
  - perform the synchronization request with the last sync-token **=> minimum bandwith usage**
  - download all _new_ objects returned by the request (if there is any)  **=> nearly zero cpu usage**
  - re-download all _changed_ objects returned by the request (if there is any) **=> nearly zero cpu usage**
  - delete all _deleted_ objects returned by the request (if there is any) **=> nearly zero cpu usage**
- while(true);

It could be a good idea to implement this feature.

Unrud:
Fixes #306.

### DIFF
--- a/config
+++ b/config
@@ -110,6 +110,9 @@
 # power fails!
 #filesystem_fsync = True
 
+# Delete sync token that are older (seconds)
+#max_sync_token_age = 2592000
+
 # Close the lock file when no more clients are waiting.
 # This option is not very useful in general, but on Windows files that are
 # opened cannot be deleted.
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -821,7 +821,8 @@ class Application:
             else:
                 collection = item.collection
             headers = {"Content-Type": "text/xml; charset=%s" % self.encoding}
-            xml_answer = xmlutils.report(
+            status, xml_answer = xmlutils.report(
                 base_prefix, path, xml_content, collection)
-            return (client.MULTI_STATUS, headers,
-                    self._write_xml_content(xml_answer))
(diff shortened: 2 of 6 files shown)


Unrud:
I've added tests for the following problems and fixed them:

  * If an item is created and deleted between two sync-requests, the sync-report must contain the deletion.
  * If an item is changed and the change is undone between two sync-requests, the sync-report must contain the change.

Unrud:
Rebased, fixed all flake8 warnings, simplified the commits and clarified some comments.

PR #568, opened 2017-03-12
partial from pbiering-testing

pbiering:
- debug_filter (optional)
- exception logging (optional)
- performance logging (optional)
- minor loglevel adjustments
- catch some error cases
- disable announcement of sync-token (still not implemented)
- example (inactive) configuration for dedicated debug log

pending: will be PRs on-top after migration:
- "request_token" + some other minor tweaks
- caching code for items+props (for 2.1 or 3.x)

### DIFF
--- a/config
+++ b/config
@@ -120,14 +120,37 @@
 # http://docs.python.org/library/logging.config.html
 #config = /etc/radicale/logging
 
-# Set the default logging level to debug
+# Set the default logging level to debug (this enables also request/response logging)
 #debug = False
+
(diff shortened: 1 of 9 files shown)


Unrud:
I think that most of the stuff in this PR got handled. The debug filtering is overdone in my opinion. Instead I tried to log everything that is important to the user as ERROR or WARNING with useful error messages. Stack traces and other stuff that pollutes the log gets only included when debugging is enabled.

Unrud:
I've added the log file rotation example to the documentation.

PR #569, opened 2017-03-13
Make Radicale fast

Unrud:
Big calendars/addressbooks are really slow and requests require a lot of memory. The main reason is vobject. Parsing and serializing with it is slow and it's a memory hog.
This PR reduces the usage of vobject by caching serialization results and metadata for pre-filtering on disk. The memory usage gets further reduced by only keeping one item in memory at a time.

All commits until ``Cache metadata when storage is read-only`` are general improvements. They fix some small bugs, reduce the memory usage by not loading all items at once into memory and fix the slowness that was introduced by collision detection for file system paths.
The following commits improve speed by:
  * Caching the serialization of the whole collection on disk.
  * Caching the serialization of items and metadata (tag, start and end time) for pre-filtering on disk.
  * Caching the etag and metadata of the collection for the duration of the request in memory, if the access is read-only.

## Benchmark
The tests are performed on a calendar and addressbook with 10,000 files. The approximate memory usage of the requests is noted in brackets.

### Calendar
```
Type:   PROPFIND_getetag (initialize cache)
Before: 15.5 seconds (138,156 kB)
After:  17.8 seconds (    372 kB)

Type:   PROPFIND_getetag
Before: 15.3 seconds (136,604 kB)
After:   0.5 seconds (      0 kB)

Type:   REPORT_calendar-query_getetag
Before: 16.8 seconds (154,180 kB)
After:   1.0 seconds ( 21,964 kB)
PR #586: 1.0 seconds ( 15,356 kB)

Type:   REPORT_calendar-multiget_getetag
Before: 16.4 seconds ( 28,248 kB)
After:   2.1 seconds ( 28,660 kB)
PR #586: 2.2 seconds ( 19,852 kB)

Type:   REPORT_calendar-query_filter-all_getetag
Before: 18.3 seconds (154,448 kB)
After:   1.1 seconds ( 22,332 kB)
PR #586: 1.1 seconds ( 15,092 kB)

Type:   REPORT_calendar-query_filter-none_getetag
Before: 11.0 seconds (133,580 kB)
After:   0.5 seconds (      0 kB)

Type:   GET
Before: 30.5 seconds (138,756 kB)
After:   1.2 seconds (  2,740 kB)
```

### Addressbook
```
Type:   PROPFIND_getetag (initialize cache)
Before:  8.0 seconds ( 69,732 kB)
After:   9.4 seconds (    284 kB)

Type:   PROPFIND_getetag
Before:  8.0 seconds ( 70,020 kB)
After:   0.5 seconds (      0 kB)

Type:   GET
Before: 16.0 seconds ( 70,012 kB)
After:   1.1 seconds (  3,692 kB)
```
**Note:** In all "Before" tests the collision checking for file names was disabled, because it slowed down the requests significantly and made the results less comparable.

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -870,6 +870,33 @@ class Collection(BaseCollection):
                 changes.append(href)
         return token, changes
 
+    @classmethod
+    def _clean_cache(cls, folder, names, max_age=None):
+        # Delete all ``names`` in ``folder`` that are older than ``max_age``.
+        age_limit = time.time() - max_age if max_age is not None else None
+        modified = False
+        for name in names:
+            if not is_safe_filesystem_path_component(name):
+                continue
+            if age_limit is not None:
+                try:
+                    # Race: Another process might have deleted the file.
+                    mtime = os.path.getmtime(os.path.join(folder, name))
+                except FileNotFoundError:
+                    continue
+                if mtime > age_limit:
+                    continue
+            cls.logger.debug("Found expired item in cache: %s", name)
+            # Race: Another process might have deleted or locked the
+            # file.
+            try:
+                os.remove(os.path.join(folder, name))
+            except (FileNotFoundError, PermissionError):
+                continue
+            modified = True
+        if modified:
+            cls._sync_directory(folder)
+
     def list(self):
         for href in scandir(self._filesystem_path, only_files=True):
             if not is_safe_filesystem_path_component(href):
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -329,14 +329,17 @@ class BaseCollection:
         raise NotImplementedError
 
     def get_multi(self, hrefs):
-        """Fetch multiple items. Duplicate hrefs must be ignored.
+        """Fetch multiple items.
 
         Functionally similar to ``get``, but might bring performance benefits
-        on some storages when used cleverly.
+        on some storages when used cleverly. It's not required to return the
+        requested items in the correct order. Duplicated hrefs can be ignored.
+
+        Returns tuples with the href and the item or None if the item doesn't
+        exist.
 
         """
-        for href in set(hrefs):
-            yield self.get(href)
+        return map(lambda href: (href, self.get(href)), hrefs)
 
     def pre_filtered_list(self, filters):
         """List collection items with optional pre filtering.
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -858,8 +858,6 @@ class Collection(BaseCollection):
                 yield href
 
     def get(self, href):
-        if not href:
-            return None
         if not is_safe_filesystem_path_component(href):
             self.logger.debug("Can't translate name %r safely to filesystem "
                               "in %r", href, self.path)
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -1113,15 +1113,43 @@ class Collection(BaseCollection):
     def serialize(self):
         # serialize collection
         if self.get_meta("tag") == "VCALENDAR":
-            collection = vobject.iCalendar()
+            vtimezones = ""
+            included_tzids = set()
+            vtimezone = []
+            tzid = None
+            components = ""
             for item in self.get_all():
-                for content in ("vevent", "vtodo", "vjournal"):
-                    if content in item.contents:
-                        for item_part in getattr(item,
-                                                 "%s_list" % content):
-                            collection.add(item_part)
-                        break
-            return collection.serialize()
+                depth = 0
+                for line in item.serialize().split("\r\n"):
+                    if line.startswith("BEGIN:"):
+                        depth += 1
+                    if depth == 2 and line == "BEGIN:VTIMEZONE":
+                        vtimezone.append(line)
+                    elif vtimezone:
+                        vtimezone.append(line)
+                        if depth == 2 and line.startswith("TZID:"):
+                            tzid = line[len("TZID:"):]
+                        elif depth == 2 and line.startswith("END:"):
+                            if tzid is None or tzid not in included_tzids:
+                                if vtimezones:
+                                    vtimezones += "\r\n"
+                                vtimezones += "\r\n".join(vtimezone)
+                                included_tzids.add(tzid)
+                            vtimezone.clear()
+                            tzid = None
+                    elif depth >= 2:
+                        if components:
+                            components += "\r\n"
+                        components += line
+                    if line.startswith("END:"):
+                        depth -= 1
+            return "\r\n".join(filter(bool, (
+                "BEGIN:VCALENDAR",
+                "VERSION:2.0",
+                "PRODID:-//PYVOBJECT//NONSGML Version 1//EN",
+                vtimezones,
+                components,
+                "END:VCALENDAR")))
         elif self.get_meta("tag") == "VADDRESSBOOK":
             return "".join(map(lambda x: x.serialize(), self.get_all()))
         return ""
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -1124,24 +1124,29 @@ class Collection(BaseCollection):
                 for line in item.serialize().split("\r\n"):
                     if line.startswith("BEGIN:"):
                         depth += 1
-                    if depth == 2 and line == "BEGIN:VTIMEZONE":
-                        vtimezone.append(line)
-                    elif vtimezone:
-                        vtimezone.append(line)
-                        if depth == 2 and line.startswith("TZID:"):
-                            tzid = line[len("TZID:"):]
-                        elif depth == 2 and line.startswith("END:"):
-                            if tzid is None or tzid not in included_tzids:
-                                if vtimezones:
-                                    vtimezones += "\r\n"
-                                vtimezones += "\r\n".join(vtimezone)
-                                included_tzids.add(tzid)
-                            vtimezone.clear()
-                            tzid = None
-                    elif depth >= 2:
-                        if components:
-                            components += "\r\n"
-                        components += line
+                    if depth == 1 and line == "BEGIN:VCALENDAR":
+                        in_vcalendar = True
+                    elif in_vcalendar:
+                        if depth == 1 and line.startswith("END:"):
+                            in_vcalendar = False
+                        if depth == 2 and line == "BEGIN:VTIMEZONE":
+                            vtimezone.append(line)
+                        elif vtimezone:
+                            vtimezone.append(line)
+                            if depth == 2 and line.startswith("TZID:"):
+                                tzid = line[len("TZID:"):]
+                            elif depth == 2 and line.startswith("END:"):
+                                if tzid is None or tzid not in included_tzids:
+                                    if vtimezones:
+                                        vtimezones += "\r\n"
+                                    vtimezones += "\r\n".join(vtimezone)
+                                    included_tzids.add(tzid)
+                                vtimezone.clear()
+                                tzid = None
+                        elif depth >= 2:
+                            if components:
+                                components += "\r\n"
+                            components += line
                     if line.startswith("END:"):
                         depth -= 1
             return "\r\n".join(filter(bool, (
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -940,11 +940,10 @@ class Collection(BaseCollection):
 
     @property
     def last_modified(self):
-        relevant_files = [self._filesystem_path] + [
-            path_to_filesystem(self._filesystem_path, href)
-            for href in self.list()]
-        if os.path.exists(self._props_path):
-            relevant_files.append(self._props_path)
+        relevant_files = chain(
+            (self._filesystem_path,),
+            (self._props_path,) if os.path.exists(self._props_path) else (),
+            map(lambda x: os.path.join(self._filesystem_path, x), self.list()))
         last = max(map(os.path.getmtime, relevant_files))
         return time.strftime("%a, %d %b %Y %H:%M:%S GMT", time.gmtime(last))
 
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -641,11 +641,9 @@ class Collection(BaseCollection):
                     cls.logger.debug("Skipping collection %r in %r", href,
                                      path)
                 continue
-            child_filesystem_path = path_to_filesystem(filesystem_path, href)
-            if os.path.isdir(child_filesystem_path):
-                child_path = posixpath.join(path, href)
-                child_principal = len(attributes) == 0
-                yield cls(child_path, child_principal)
+            child_path = posixpath.join(path, href)
+            child_principal = len(attributes) == 0
+            yield cls(child_path, child_principal)
 
     @classmethod
     def create_collection(cls, href, collection=None, props=None):
@@ -826,7 +824,7 @@ class Collection(BaseCollection):
         history_folder = os.path.join(self._filesystem_path,
                                       ".Radicale.cache", "history")
         try:
-            for href in os.listdir(history_folder):
+            for href in scandir(history_folder):
                 if not is_safe_filesystem_path_component(href):
                     continue
                 if os.path.isfile(os.path.join(self._filesystem_path, href)):
@@ -868,7 +866,7 @@ class Collection(BaseCollection):
         token_name_hash = md5()
         # Find the history of all existing and deleted items
         for href, item in chain(
-                ((item.href, item) for item in self.pre_filtered_list(())),
+                ((item.href, item) for item in self.get_all()),
                 ((href, None) for href in self._get_deleted_history_hrefs())):
             history_etag = self._update_history_etag(href, item)
             state[href] = history_etag
@@ -936,33 +934,6 @@ class Collection(BaseCollection):
                 changes.append(href)
         return token, changes
 
-    @classmethod
-    def _clean_cache(cls, folder, names, max_age=None):
-        # Delete all ``names`` in ``folder`` that are older than ``max_age``.
-        age_limit = time.time() - max_age if max_age is not None else None
-        modified = False
-        for name in names:
-            if not is_safe_filesystem_path_component(name):
-                continue
-            if age_limit is not None:
-                try:
-                    # Race: Another process might have deleted the file.
-                    mtime = os.path.getmtime(os.path.join(folder, name))
-                except FileNotFoundError:
-                    continue
-                if mtime > age_limit:
-                    continue
-            cls.logger.debug("Found expired item in cache: %s", name)
-            # Race: Another process might have deleted or locked the
-            # file.
-            try:
-                os.remove(os.path.join(folder, name))
-            except (FileNotFoundError, PermissionError):
-                continue
-            modified = True
-        if modified:
-            cls._sync_directory(folder)
-
     def list(self):
         for href in scandir(self._filesystem_path, only_files=True):
             if not is_safe_filesystem_path_component(href):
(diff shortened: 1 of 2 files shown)


pbiering:
sounds interesting and probably superseed my "simple" in-memory caching, yours is persistent and survives Radicale restarts.

Unrud:
I've implemented prefiltering for the multifilesystem backend and all collection items are now verified by vobject. For this I added a on-disk cache that stores the etags of verified items and some extracted parameter for prefiltering. Items that are not in the cache are always parsed and serialized before they are returned by **Collection.get**.

Prefiltering takes about 1 second to filter out 10,000 items.

I've added some helper functions for prefiltering to xmlutils. They convert the complicated XML filters to a tag and time range (POSIX timestamps as ints) and can find the tag and an enclosing time range for vobject_items. 

> superseed my "simple" in-memory caching

I did some experiments. I loaded a collection (with simple VEVENTs) that required ~3.4 MiB of space on disk. Caching it in memory required an additional ~128 MiB of RAM.

Unrud:
I changed the item cache to only keep one cache entry per item. Entries for deleted items get cleaned up automatically.

Unrud:
Rebased, fixed all flake8 warnings, simplified and reordered the commits and added some comments.

All commits until ``Cache metadata when storage is read-only`` are general improvements. They fix some small, reduce the RAM usage by not loading all items at once into memory and fix the slowness that was introduced by collision detection for file system paths.
The following commits improve speed by:
  * Caching the serialization of the whole collection on disk.
  * Caching the serialization of items and metadata (tag, start and end time) for pre-filtering on disk.
  * Caching the etag and metadata of the collection for the duration of the request in memory, if the access is read-only.

Unrud:
I ran the benchmark again, to incorporate the new changes. This time I also included the approximate RAM usage of the requests. The tests are performed on a calendar and addressbook with 10,000 files.

## Benchmark
### Calendar
```
Type:   PROPFIND_getetag (initialize cache)
Before: 15.5 seconds (138,156 kB)
After:  17.8 seconds (    372 kB)

Type:   PROPFIND_getetag
Before: 15.3 seconds (136,604 kB)
After:   0.5 seconds (      0 kB)

Type:   REPORT_calendar-query_getetag
Before: 16.8 seconds (154,180 kB)
After:   1.0 seconds ( 21,964 kB)

Type:   REPORT_calendar-multiget_getetag
Before: 16.4 seconds ( 28,248 kB)
After:   2.1 seconds ( 28,660 kB)

Type:   REPORT_calendar-query_filter-all_getetag
Before: 18.3 seconds (154,448 kB)
After:  10.2 seconds ( 22,332 kB)

Type:   REPORT_calendar-query_filter-none_getetag
Before: 11.0 seconds (133,580 kB)
After:   0.5 seconds (      0 kB)

Type:   GET (initialize cache)
Before: 30.4 seconds (138,752 kB)
After:  16.3 seconds (102,228 kB)

Type:   GET
Before: 30.5 seconds (138,756 kB)
After:   0.6 seconds (  6,860 kB)
```

### Addressbook
```
Type:   PROPFIND_getetag (initialize cache)
Before:  8.0 seconds ( 69,732 kB)
After:   9.4 seconds (    284 kB)

Type:   PROPFIND_getetag
Before:  8.0 seconds ( 70,020 kB)
After:   0.5 seconds (      0 kB)

Type:   GET (initialize cache)
Before: 16.0 seconds ( 70,012 kB)
After:   1.1 seconds (  3,836 kB)

Type:   GET
Before: 15.8 seconds ( 69,460 kB)
After:   0.6 seconds (  5,012 kB)
```
**Note:** In all "Before" tests the collision checking for file names was disabled, because it slowed down the requests significantly and made the results less comparable.

Unrud:
I implemented the serialization of calendars without vobject. The calendar components are merged manually. This improves the time of GET requests from ~30 seconds (~100 MiB of memory) to ~1 second (~5 MiB of memory). Consequentially I removed the cache for the serialization of collections, because it's not needed anymore.

Filtering of items with vobject in ``xmlutils.py`` can now be skipped if the storage backend indicates that it already matched all ``filters`` in prefiltering. This is used to optimize requests with simple filters, that only match the tag and a big time range (like filtering all events from the last three months).
At least DAVdroid and InfCloud only make use of simple filters. This improves the time for the requests from ~10 seconds to ~1 second.

vobject is now only used for one-time input validation and for complex filtering.
I've updated the benchmark in the first post and also included results from #586.

I've created a fork of Radicale at https://github.com/Unrud/Radicale where I merged all my open PRs.

jvoisin:
Nice!
What is needed to get this merged @liZe ?

radicale/storage.py
```diff
@@ -687,53 +862,87 @@ def delete(self, href=None):
...
+                        vtimezone.append(line)
+                        if depth == 2 and line.startswith("TZID:"):
+                            tzid = line[len("TZID:"):]
+                        elif depth == 2 and line.startswith("END:"):
+                            if tzid is None or tzid not in included_tzids:
+                                if vtimezones:
+                                    vtimezones += "\r\n"
+                                vtimezones += "\r\n".join(vtimezone)
+                                included_tzids.add(tzid)
+                            vtimezone.clear()
```
htgoebel: Why not simply `vtimezone = []`, which would be more obfious?

radicale/storage.py
```diff
@@ -687,53 +862,87 @@ def delete(self, href=None):
...
-                        for item_part in getattr(item, "%s_list" % content):
-                            collection.add(item_part)
-                        break
-            return collection.serialize()
+            vtimezones = ""
+            included_tzids = set()
+            vtimezone = []
+            tzid = None
+            components = ""
+            for item in self.get_all():
```
htgoebel: Please add a comment what this code does and why.

radicale/storage.py
```diff
@@ -687,53 +862,87 @@ def delete(self, href=None):
...
+                depth = 0
+                for line in item.serialize().split("\r\n"):
+                    if line.startswith("BEGIN:"):
+                        depth += 1
+                    if depth == 2 and line == "BEGIN:VTIMEZONE":
+                        vtimezone.append(line)
+                    elif vtimezone:
+                        vtimezone.append(line)
+                        if depth == 2 and line.startswith("TZID:"):
+                            tzid = line[len("TZID:"):]
```
htgoebel: Use a constant here : `line = line[5:]`

radicale/storage.py
```diff
@@ -687,53 +862,87 @@ def delete(self, href=None):
...
+            for item in self.get_all():
+                depth = 0
+                for line in item.serialize().split("\r\n"):
+                    if line.startswith("BEGIN:"):
+                        depth += 1
+                    if depth == 2 and line == "BEGIN:VTIMEZONE":
+                        vtimezone.append(line)
+                    elif vtimezone:
+                        vtimezone.append(line)
+                        if depth == 2 and line.startswith("TZID:"):
```
htgoebel: This would also match if `TZID:` is part of another nested structure. Is this indented?

radicale/storage.py
```diff
@@ -687,53 +862,87 @@ def delete(self, href=None):
...
         return ""
 
+    @property
+    def etag(self):
+        # reuse cached value if the storage is read-only
+        if self._writer or self._etag is None:
+            etag = md5()
+            for item in self.get_all():
+                etag.update((item.href + "/" + item.etag).encode("utf-8"))
+            self._etag = '"%s"' % etag.hexdigest()
```
htgoebel: Using string-concatenation (`+`) is more obvious and should be as fast.

radicale/xmlutils.py
```diff
@@ -207,11 +209,9 @@ def _prop_match(item, filter_):
 
 
 def _time_range_match(vobject_item, filter_, child_name):
-    """Check whether the ``item`` matches the time-range ``filter_``.
+    """Check whether the the component/property ``child_name`` of
```
htgoebel: typo

radicale/xmlutils.py
```diff
@@ -226,14 +226,58 @@ def _time_range_match(vobject_item, filter_, child_name):
...
+    operation is cancelled.
+
+    See rfc4791-9.9.
+
+    """
+
+    DAY = timedelta(days=1)
+    SECOND = timedelta(seconds=1)
+    DATETIME_MIN = datetime.min.replace(tzinfo=timezone.utc)
+    DATETIME_MAX = datetime.max.replace(tzinfo=timezone.utc)
```
htgoebel: These are constant values and should be global.

radicale/xmlutils.py
```diff
@@ -426,6 +493,107 @@ def _param_filter_match(vobject_item, filter_, parent_name):
...
+    a string or None (match all) and ``start`` and ``end`` are POSIX
+    timestamps (as int). ``simple`` is a bool that indicates that ``filters``
+    and the simplified condition are identical.
+
+    """
+
+    TIMESTAMP_MIN = math.floor(
+        datetime.min.replace(tzinfo=timezone.utc).timestamp())
+    TIMESTAMP_MAX = math.ceil(
+        datetime.max.replace(tzinfo=timezone.utc).timestamp())
```
htgoebel: These are constants and should be global

radicale/xmlutils.py
```diff
@@ -426,6 +493,107 @@ def _param_filter_match(vobject_item, filter_, parent_name):
...
+
+    Returns a tuple (``tag``, ``start``, ``end``) where ``tag`` is a string
+    and ``start`` and ``end`` are POSIX timestamps (as int).
+
+    This is intened to be used for matching against simplified prefilters.
+
+    """
+
+    DATETIME_MIN = datetime.min.replace(tzinfo=timezone.utc)
+    DATETIME_MAX = datetime.max.replace(tzinfo=timezone.utc)
```
htgoebel: dito.

radicale/storage.py
```diff
@@ -207,19 +237,42 @@ def __init__(self, path):
 
 
 class Item:
-    def __init__(self, collection, item, href, last_modified=None):
+    def __init__(self, collection, href, last_modified=None,
+                 text=None, item=None, etag=None):
```
Unrud: > Please add a doc-string

Yes.

> The interface changed.

Totally missed that. I also changed the behavior of ``pre_filtered_list``.

> This should go into the release notes.

Or I reorder the parameters and make it backwards compatible. It depends, if this PR gets merged before the 2.0.0 release.

radicale/storage.py
```diff
@@ -207,19 +237,42 @@ def __init__(self, path):
...
+        if text is None and item is None:
```
Unrud: ``text`` is just the raw data from the storage. It will fail as soon as it's parsed with **vobject**, like any other invalid data. I don't think that we should introduce a special case for one kind of invalid data. That would also mean, that we have to duplicate the logging.

radicale/storage.py
```diff
@@ -207,19 +237,42 @@ def __init__(self, path):
...
+            raise ValueError("text and item are not set")
```
Unrud: Yes.

radicale/storage.py
```diff
@@ -304,23 +357,40 @@ def get(self, href):
...
+
+        Functionally similar to ``get``, but might bring performance benefits
+        on some storages when used cleverly. It's not required to return the
+        requested items in the correct order. Duplicated hrefs can be ignored.
+
+        Returns tuples with the href and the item or None if the item doesn't
+        exist.
+
+        """
+        return map(lambda href: (href, self.get(href)), hrefs)
```
Unrud: No, that would defeat the main purpose of this PR. A generator expression will load everything into memory at once. Avoiding generator expressions is the main reason, why this PR reduces the memory usage significantly.

radicale/storage.py
```diff
@@ -304,23 +357,40 @@ def get(self, href):
...
+    def get_all(self):
+        """Fetch all items.
 
         Functionally similar to ``get``, but might bring performance benefits
         on some storages when used cleverly.
 
         """
-        for href in set(hrefs):
-            yield self.get(href)
+        return map(self.get, self.list())
```
Unrud: No, see comment about generator expressions.

radicale/storage.py
```diff
@@ -304,23 +357,40 @@ def get(self, href):
...
         the filters and this implementation.
+
+        Returns tuples in the form ``(item, filters_matched)``.
+        ``filters_matched`` is a bool that indicates if ``filters`` are fully
+        matched.
+
         This returns all event by default
         """
-        return [self.get(href) for href in self.list()]
+        return map(lambda item: (item, False), self.get_all())
```
Unrud: No, see comment about generator expressions.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
+                if mtime > age_limit:
+                    continue
+            cls.logger.debug("Found expired item in cache: %s", name)
+            # Race: Another process might have deleted or locked the
+            # file.
+            try:
+                os.remove(os.path.join(folder, name))
+            except (FileNotFoundError, PermissionError):
+                continue
+            modified = True
```
Unrud: > Shorter and more focused on on what should be done here

It's not equivalent. The file is not removed if ``age_limit`` is ``None``.

> compute path only once for speed

Maybe, but it shouldn't make a measurable difference.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
+                # Race: Other processes might have created and locked the
+                # file.
+                with self._atomic_write(os.path.join(cache_folder, href),
+                                        "wb") as f:
+                    pickle.dump((input_hash, cetag, ctext,
+                                 ctag, cstart, cend), f)
+            except PermissionError:
+                pass
+            # Clean cache entries (max once per request)
+            if not self._item_cache_cleaned:
```
Unrud: This only happens once, after new items are uploaded to the storage. I put it here and not into ``def upload(...)`` because the user can also edit the file system storage by hand.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
+                self._item_cache_cleaned = True
+
+                def is_invalid_href(href):
+                    return not os.path.isfile(
+                        os.path.join(self._filesystem_path, href))
+
+                self._clean_cache(cache_folder, filter(
+                    is_invalid_href, scandir(cache_folder)))
```
Unrud: No, see comment about generator expressions. Although in this case it's a bit overkill, this doesn't save any significant amount of memory. Directory contents are usually not that big.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
             "%a, %d %b %Y %H:%M:%S GMT",
             time.gmtime(os.path.getmtime(path)))
-        try:
-            item = vobject.readOne(text)
-        except Exception:
-            self.logger.error("Failed to parse component: %s", href)
-            raise
-        return Item(self, item, href, last_modified)
+        return Item(self, href, last_modified=last_modified, etag=cetag,
+                    text=ctext, item=vobject_item), (ctag, cstart, cend)
```
Unrud: You can't return ``vobject_item`` directly. The ``text`` is the raw text from the file system. You have to serialize it again, to normalize the text with **vobject**. The user might have edited the storage in the file system by hand, it might not be normalized.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
+
+    def get_multi(self, hrefs):
+        # It's faster to check for file name collissions here, because
+        # we only need to call os.listdir once.
+        files = None
+        for href in hrefs:
+            if files is None:
+                # Only list dir when hrefs is not empty
```
Unrud: That is alright. If the ``hrefs`` iterator doesn't contain anything, the for-loop is never executed.
Therefore, we only list the directory content if the ``hrefs`` iterator is not empty. That's the whole purpose of this.

Unfortunately it's somewhat complicated to check if the iterator is empty, before the for-loop.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
+                self.logger.debug(
+                    "Can't translate name safely to filesystem: %s", href)
+                yield (href, None)
+            else:
+                yield (href, self.get(href, verify_href=False))
+
+    def get_all(self):
+        # We don't need to check for collissions, because the the file names
+        # are from os.listdir.
+        return map(lambda x: self.get(x, verify_href=False), self.list())
```
Unrud: No, see comment about generator expressions.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
+
+    def pre_filtered_list(self, filters):
+        tag, start, end, simple = xmlutils.simplify_prefilters(filters)
+        if not tag:
+            # no filter
+            yield from map(lambda item: (item, simple), self.get_all())
+            return
```
Unrud: I tried to prevent deep indentations.

radicale/storage.py
```diff
@@ -621,43 +689,150 @@ def move(cls, item, to_collection, to_href):
...
+        for item, (itag, istart, iend) in map(
+                lambda x: self._get_with_metadata(x, verify_href=False),
+                self.list()):
```
Unrud: No, see comment about generator expressions.

radicale/storage.py
```diff
@@ -687,53 +862,87 @@ def delete(self, href=None):
...
-                old_props = json.load(f)
-                old_props.update(props)
-                props = old_props
-        props = {key: value for key, value in props.items() if value}
-        with self._atomic_write(self._props_path, "w+") as f:
-            json.dump(props, f)
+        new_props = self.get_meta()
+        new_props.update(props)
+        for key in tuple(new_props.keys()):
+            if not new_props[key]:
```
Unrud: It's the same behavior as the old code. ``xmlutils`` uses it to delete entries.

radicale/storage.py
Unrud: It doesn't because of ``depth == 2``.

radicale/storage.py
Unrud: I tried to avoid magic numbers.

radicale/storage.py
Unrud: Because it uses more memory (the GC doesn't cleanup the old memory immediately) and it's slower. This makes a big difference.

radicale/storage.py
Unrud: > Using string-concatenation

I used this method, because it's also used at other places in the code. I don't think that it matters.

> should be as fast.

This is only done once, speed is not a concern.

radicale/xmlutils.py
Unrud: Yes.

radicale/xmlutils.py
Unrud: Yes.

radicale/xmlutils.py
Unrud: Yes.

radicale/xmlutils.py
Unrud: Yes.

Unrud:
@htgoebel Thanks for reviewing, I will fix the identified problems soon.

radicale/storage.py
htgoebel: Puh, this is tricky :-) I suggest adding a comment so other nitpickers keep quiet.

radicale/storage.py
htgoebel: I'm curios about this. AFAIK (and I'm quite sure about this) generator expressions (using parents `(…)`) are like iterators. Opposed to list comprehension (using brackets `[…]`) which are like lists. And iterators are not loading all data into memory at once. But I may be wrong and am really curious to learn why.

radicale/storage.py
htgoebel: I mean this case:
```
BEGIN:Rubbish
BEGIN:VTIMEZONE
TZID:
…
```

htgoebel: I mean this case (which *may* be an invalid vevent, but this *may* change in some revision of the standard)
```
BEGIN:Rubbish
BEGIN:VTIMEZONE
TZID:
…
```

radicale/storage.py
Unrud: You are right, I I mixed them up.

PR #581, opened 2017-04-22
Add git init to config hook example

jre-wine:
The example code previously required to manually create a git repo first.
Above change automates this.

### DIFF
--- a/config
+++ b/config
@@ -109,7 +109,7 @@
 #filesystem_close_lock_file = False
 
 # Command that is run after changes to storage
-# Example: git add -A && (git diff --cached --quiet || git commit -m "Changes by "%(user)s)
+# Example: ([ -d .git ] || git init) && git add -A && (git diff --cached --quiet || git commit -m "Changes by "%(user)s)
 #hook =
 
 


Unrud:
I would rather add a comment instead of integrating it into the command line.
The comment could also mention additional useful setup steps, like adding ``.Radicale.tmp-*`` to ``.gitignore``.

jre-wine:
@Unrud that works too of course. I just propose to have all git related instructions in the same place (either in the config file or in separate documentation).
I'm using my proposed config since half a year. So far I didn't notice any tmp files, although I have no .gitignore file.

Unrud:
> I'm using my proposed config since half a year. So far I didn't notice any tmp files, although I have no .gitignore file.

The temporary files are only left behind in case of some errors. E.g.:

  * A new item is uploaded and it can't be written to disk completely, because not enough free space is available.
  * Radicale or the whole computer crashes during atomic creation of files.

liZe:
Thank you! The command line is OK for me and will help the crazy-cut-and-paste users, it's simple enough to be understood by the ones who want to spend more time understanding what they're doing.

PR #587, opened 2017-05-07
Capture log messages during tests

Unrud:
The log messages help finding problems with failed tests.

### DIFF
--- a/radicale/tests/__init__.py
+++ b/radicale/tests/__init__.py
@@ -19,15 +19,25 @@ Tests for Radicale.
 
 """
 
+import logging
 import os
 import sys
 from io import BytesIO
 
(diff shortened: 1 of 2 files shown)


htgoebel:
A better approach – but more work and more changes – would be to switch to pytest and use `pytest-logcatch`.

Unrud:
> A better approach

Why is it better?

htgoebel:
> Why is it better?

Because
a) it reuses code written, maintained and tested by others
b) pytest-logcatch has a lot of useful features, like a "set log-level" context manager, checking log entries on a "record" basies, supports setting the log format via setup.cfg or command-line arguments, etc.

PR #598, opened 2017-05-22
Fix: SCRIPT_NAME is not stripped from collections path.

htgoebel:
This is required if radicale is served e.g. via http://server/radicale/


htgoebel:
http://flask.pocoo.org/snippets/35/ has a different approach: it removes script name from environ['PATH_INFO'], while my code on removes it from the variable `path`.
```
            path_info = environ['PATH_INFO']
            if path_info.startswith(script_name):
                environ['PATH_INFO'] = path_info[len(script_name):]
```

liZe:
Thanks for this pull request!

I've cherry-picked the equivalent 577d931, this should be fixed now.

PR #600, opened 2017-05-23
Test and fix auth module. Configurable delay. Improve logging.

Unrud:
* Repair the tests for the auth module (they were not working at all).
  * Add tests for all hash methods and also test with wrong usernames and passwords.
  * Fix the SSHA method.
  * Compare in constant time.
  * Add configuration option for the delay.
  * Don't apply the delay for clients that didn't try to authenticate.

### DIFF
--- a/config
+++ b/config
@@ -78,6 +78,9 @@
 # bcrypt and md5 require the passlib library to be installed.
 #htpasswd_encryption = bcrypt
 
+# Incorrect authentication delay (seconds)
+#delay = 1
+
 
 [rights]
 
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -34,11 +34,13 @@ import itertools
 import os
 import posixpath
 import pprint
+import random
 import socket
 import socketserver
 import ssl
 import sys
 import threading
+import time
 import traceback
 import wsgiref.simple_server
 import zlib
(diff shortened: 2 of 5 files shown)


liZe:
Thanks **a lot** for this. :heart:

The situation is way better with these commits, even if I'm not sure that it fixes the security problem correctly. We probably need at least to:
- use `compare_digest` for logins, and
- test the password even if the logins don't match.

@Unrud @jvoisin What do you think of that?

jvoisin:
I wholeheartedly agree :)

Unrud:
> I'm not sure that it fixes the security problem correctly

I ignored the user names, because the timing attacks seems pretty unrealistic to me. To pull this off you would have to learn the internal state of Python's PRNG from delays with network jitter. Next you would have to compare the response times of different user names (again with network jitter). To do all of this you have less than 20 tries per second (because of the delay and maximal number of parallel connections.)

You do all of this, just to learn if a user name exists or not.

liZe:
> I ignored the user names, because the timing attacks seems pretty unrealistic to me.

Indeed. My deep feeling about that is that it's bullsh^H useless. But some some smart people sent me private mails with long security papers that convinced me to try to fix this issue correctly.

jvoisin:
@Unrud no need to play with the PRNG, just launch your attack over a long period of time: if you're not monitoring your instance, odds are that you won't see my 20 requests per second during 6 months ;)
There was a [paper]( https://www.blackhat.com/docs/us-15/materials/us-15-Morgan-Web-Timing-Attacks-Made-Practical-wp.pdf ) two years ago at Blackhat US about this, and attacks are only getting better.

An interesting threat model would be hidden-services, where people are dumb enough to use their everyday-nick to login on it.

But yeah, I agree, it's completely overkill. But it's cheap, doesn't bloat the code, so why not.

PR #604, opened 2017-05-24
Disable logging config by default

Unrud:
Radicale always tries to load the system-wide configuration file. To turn this off, the logging-config option has to be added to all configuration files and command line arguments. It's easier to disable it by default and only enable it once in the system-wide config file.


liZe:
No problem, it should have been done before.

PR #605, opened 2017-05-24
Move "realm" configuration from "server" to "auth" section

Unrud:
All configuration options in the ``server`` section should only affect the standalone server.


Unrud:
Closed, because there is also the ``max_connection_length`` option.

PR #608, opened 2017-05-24
Fail if config file is not found

Unrud:
If a configuration file is passed with a command line argument and the file is not found, Radicale shows a warning and continues with the default configuration.
There is no reason for doing this, Radicale should just fail.

Instead, this PR allows passing an empty string like ``--config ""``. Radicale will use the default configuration in this case, without trying to load the configuration from the common paths. Previously you had to do specify a path that doesn't exist like ``--config /does/not/exist``, which looks ugly and showed a warning message.

### DIFF
--- a/radicale/__main__.py
+++ b/radicale/__main__.py
@@ -74,9 +74,12 @@ def run():
                 group.add_argument(*args, **kwargs)
 
     args = parser.parse_args()
-    if args.config:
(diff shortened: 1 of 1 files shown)


liZe:
> There is no reason for doing this, Radicale should just fail.

You're right.

Issue #603, opened 2017-05-24
Report uri causing errors in the logfile


htgoebel:
If an server error occurs while processing some data (like in #602), the uri causing the error should be written to the log-file, using level ERROR. This would easy debugging both the server and the data.

htgoebel:
FYI: I implemented this, pull-request to come.

Unrud:
Fixed in development version

Issue #601, opened 2017-05-23
Gracefully report data validation errors


htgoebel:
When migrating old data, I encountered quite some problems with wrong data, e.g. calendar entries containing lines unparsable by vObject, or vCards failing validation.

In this case the server's request handler crashes (see below) and the concrete error message shows up only in the log file. The clients receives status code 500, indicating a server error, which is wrong.

It would be very helpful if radicale would catch these errors and report them to the client (using status code 422, see [RFC 4918 section11.2](https://tools.ietf.org/html/rfc4918#section-11.2).

```
An exception occurred during request:
Traceback (most recent call last):
File "/usr/lib/python3.4/wsgiref/handlers.py", line 137, in run
self.result = application(self.environ, self.start_response)
File "/usr/local/lib/python3.4/dist-packages/radicale/__init__.py", line 414, in __call__
environ, base_prefix, path, user)
File "/usr/local/lib/python3.4/dist-packages/radicale/__init__.py", line 665, in do_PUT
items = list(vobject.readComponents(content or ""))
File "/usr/local/lib/python3.4/dist-packages/vobject/base.py", line 1090, in readComponents
vline = textLineToContentLine(line, n)
File "/usr/local/lib/python3.4/dist-packages/vobject/base.py", line 920, in textLineToContentLine
return ContentLine(*parseLine(text, n), **{'encoded': True,
File "/usr/local/lib/python3.4/dist-packages/vobject/base.py", line 808, in parseLine
raise ParseError("Failed to parse line: {0!s}".format(line), lineNumber)
vobject.base.ParseError: At line 24: Failed to parse line: <HTML>
```

htgoebel:
FYI: I implemented this, pull-request to come.

Unrud:
Bad request responses are send to clients now instead of server error. The error causing the problem gets logged with level WARNING.

PR #609, opened 2017-05-26
Improve logging

htgoebel:
Closes
* Gracefully report data error using status code 422 (issue #601)
* Report uri causing errors with level ERROR (issue #603)
*  	Log files causing errors when filtering events (no issue filed for this)

### DIFF
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -437,6 +437,9 @@ class Application:
                 return response(*UNPROCESSABLE_ENTITY(e))
             except xmlutils.ET.ParseError as e:
                 return response(*UNPROCESSABLE_ENTITY("Invalid XML: %s" % e))
+            except:
+                self.logger.error("Error processing %(PATH_INFO)s:", environ)
+                raise
             if (status, headers, answer) == NOT_ALLOWED:
                 self.logger.info("Access denied for %s",
                                  "'%s'" % user if user else "anonymous user")
--- a/radicale/__init__.py
+++ b/radicale/__init__.py
@@ -2,6 +2,7 @@
 # Copyright © 2008 Nicolas Kandel
 # Copyright © 2008 Pascal Halter
 # Copyright © 2008-2017 Guillaume Ayoub
+# Copyright © 2017 Hartmut Goebel
 #
 # This library is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -77,6 +78,13 @@ REMOTE_DESTINATION = (
 DIRECTORY_LISTING = (
     client.FORBIDDEN, (("Content-Type", "text/plain"),),
     "Directory listings are not supported.")
+
+
+def UNPROCESSABLE_ENTITY(msg):
+    return (client.UNPROCESSABLE_ENTITY,
+            (("Content-Type", "text/plain"),),
+            "Error processing the contained data: %s" % msg)
+
 
 DAV_HEADERS = "1, 2, 3, calendar-access, addressbook, extended-mkcol"
 
@@ -425,6 +433,10 @@ class Application:
                     environ, base_prefix, path, user)
             except socket.timeout:
                 return response(*REQUEST_TIMEOUT)
+            except vobject.base.VObjectError as e:
+                return response(*UNPROCESSABLE_ENTITY(e))
+            except xmlutils.ET.ParseError as e:
+                return response(*UNPROCESSABLE_ENTITY("Invalid XML: %s" % e))
             if (status, headers, answer) == NOT_ALLOWED:
                 self.logger.info("Access denied for %s",
                                  "'%s'" % user if user else "anonymous user")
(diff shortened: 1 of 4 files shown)


htgoebel:
I fixed the flak8 issues now. Sorry for the noise of intermediate commits, I confused my branches.

htgoebel:
Rebased on top of release 2.0

radicale/__init__.py
```diff
@@ -425,6 +433,13 @@ def response(status, headers=(), answer=None):
                     environ, base_prefix, path, user)
             except socket.timeout:
                 return response(*REQUEST_TIMEOUT)
+            except vobject.base.VObjectError as e:
+                return response(*UNPROCESSABLE_ENTITY(e))
```
Unrud: I don't think that 422 is the right error. It's for semantic errors, 400 should be more appropriate.

Why didn't you put it into do_PUT, where uploaded data from the client is parsed? Here it catches all errors from vobject, no matter where they happen.

radicale/__init__.py
```diff
@@ -425,6 +433,13 @@ def response(status, headers=(), answer=None):
...
+            except xmlutils.ET.ParseError as e:
+                return response(*UNPROCESSABLE_ENTITY("Invalid XML: %s" % e))
```
Unrud: dito.

radicale/__init__.py
```diff
@@ -425,6 +433,13 @@ def response(status, headers=(), answer=None):
...
+            except:
```
Unrud: This also catches ``KeyboardInterrupt``.

radicale/__init__.py
htgoebel: IMO 422 is the correct code, see the explanation in [RFC 4918 section11.2](https://tools.ietf.org/html/rfc4918#section-11.2) – which is exactly describing this case.

I'm not putting it into `do_PUT` because here is where exceptions in requests are handled. And if (for whatever reason) vObjects are parsed in another method, this is also handled .

radicale/__init__.py
htgoebel: XML parsing errors can occur in any method  which parses XML, e.g. do_REPORT.

radicale/__init__.py
htgoebel: Hmm, I don't think a thread will ever get a KeyboardInterrupt, but I can handle this case, if you like.

radicale/__init__.py
Unrud: From the RFC:
> syntax of the request entity is correct

This also catches syntax errors. 400 is more appropriate.

> And if (for whatever reason) vObjects are parsed in another method

But the 4xx errors are indicating client errors. If a vobject from the storage can't be parsed, it's not a client error.

htgoebel: > From the RFC:

>>  syntax of the request entity is correct

> This also catches syntax errors.

Indeed. But RFC 4918 continues with: "For example, this error condition may occur if an XML request body contains well-formed (i.e., syntactically correct), but semantically erroneous, XML instructions." Now replace XML by vObject and this is our case.

[Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors) says: "422: The request was well-formed but was unable to be followed due to semantic errors".

There is also a [longer discussion about this at stackoverflow](https://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data) which concludes: "Given the choice between using an error code in the original standard that explicitly doesn't cover the situation, and one from an extension that describes the situation exactly, **I would choose the latter**." (Emphasis by me)

So IMHO 422 is more appropriate then 440, since as I understand it, 422 is about the data which is transferred, while 400 is about he request. Otherwise there would have been no reason for introducing 422 for WebDAV.

> But the 4xx errors are indicating client errors. If a vobject from the storage can't be parsed, it's not a client error.

Well spotted :-) I'll move this code into "PUT" (and related) after we agreed on the error.

htgoebel:
Independent of the decision about the error-code: I'll add logging to the server's log, too. Currently the errors are reported to the client only. This inhibits spotting problems when monitoring the server logs.

Unrud:
Please take a look at #613. It overlaps with this PR.

The log output now looks something like this for invalid requests:

```
[7fe47a8ec700] WARNING: Bad PUT request on '/user/addressbook.ics/8a87451e-30c7-4d1a-906e-aeab14742f9c.ics': At line 1: Failed to parse line: Invalid
```
or
```
[7fe47a8ec700] WARNING: Bad PROPFIND request on '/': Failed to parse XML: junk after document element: line 13, column 9
```

If the files in the filesystem are damaged you get error like this:

```
[7fe47a8ec700] ERROR: An exception occurred during request: Failed to load properties of collection 'user/addressbook.ics': Expecting value: line 1 column 1 (char 0)
```
or
```
[7fe47a8ec700] ERROR: An exception occurred during request: Failed to parse item '8a87451e-30c7-4d1a-906e-aeab14742f9c.ics' in 'user/addressbook.ics'
```

PR #676, opened 2017-07-28
improving DockerFile

jaesivsm:
I tried building the DockerFile and encountered some error. Seemingly, the archive downloaded through the ADD command wasn't valid. I came accross [this article](https://www.ctl.io/developers/blog/post/dockerfile-add-vs-copy/) which implied the command can't be trusted so I replaced it with ```wget``` which is present and reliable.

I also added the ```ARG``` part so that the version can be overridden with ```docker build . --build-arg VERSION=2.1.2```.

In the end this version works for on with Debian Stretch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,10 +17,10 @@ RUN apk add --no-cache \
       build-base \
       libffi-dev
 # Install Radicale
-RUN wget --quiet https://github.com/Kozea/Radicale/archive/${VERSION}.tar.gz && \
-    tar xzf ${VERSION}.tar.gz && \
+RUN wget --quiet https://github.com/Kozea/Radicale/archive/${VERSION}.tar.gz --output-document=radicale.tar.gz && \
+    tar xzf radicale.tar.gz && \
     python3 -m pip install ./Radicale-${VERSION} && \
-    rm -r ${VERSION}.tar.gz Radicale-${VERSION}
+    rm -r radicale.tar.gz Radicale-${VERSION}
 # Persistent storage for data (Mount it somewhere on the host!)
 VOLUME /var/lib/radicale
 # Configuration data (Put the "config" file here!)
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,6 @@ FROM alpine:latest
 
 # Version of Radicale (e.g. 2.0.0)
 ARG VERSION=master
-ENV VERSION ${VERSION}
 
 # Install dependencies
 RUN apk add --no-cache \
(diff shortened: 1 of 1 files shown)


Unrud:
> Seemingly, the archive downloaded through the ADD command wasn't valid.

I think that the problem is, that Docker downloads a corrupted file. The tar command fails, because the gzip CRC is wrong. All further tries to build the image also fail, because Docker has cached the corrupted download.

How does ``RUN wget ...`` improve the situation over ``ADD``? Docker will still cache the step, if the download is corrupted.

Can you change it to something like ``RUN wget ... && tar ...``. If the download and CRC check is in the same step, Docker will not cache the downloaded file if it's corrupted.

> which implied the command can't be trusted

``ADD`` has some "magic" behavior for local files, but that's not an issue here.

> I also added the ```ARG``` part so that the version can be overridden with ```docker build . --build-arg VERSION=2.1.2```.

:thumbsup:

Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM alpine:latest
 
 # Version of Radicale (e.g. 2.0.0)
-ENV VERSION master
+ARG VERSION=master
+ENV VERSION ${VERSION}
```
Unrud: I don't think that ``ARG ...`` and ``ENV ...`` are both required. Just ``ARG ...`` should be enough.

jaesivsm: Keeping the ```ENV``` part allows to have the ```VERSION``` environment variable available during runtime. So yeah, you're right, I think it's superfluous too.

jaesivsm:
> > Seemingly, the archive downloaded through the ADD command wasn't valid.
> 
> I think that the problem is, that Docker downloads a corrupted file. The tar command fails, because the gzip CRC is wrong. All further tries to build the image also fail, because Docker has cached the corrupted download.
> 
> How does RUN wget ... improve the situation over ADD? Docker will still cache the step, if the download is corrupted.

Well I tried several build from scratch of the original image  and in my setup, unziping after ```ADD`` failed every time. I'm sure ```wget``` is the right way to go, but the build worked for me that way. 

> Can you change it to something like RUN wget ... && tar .... If the download and CRC is in the same step, Docker will not cache the downloaded file if it's corrupted. The whole step fails instead.

Sure, I'll try something more straightforward.

Dockerfile
```diff
@@ -17,10 +17,10 @@ RUN apk add --no-cache \
       build-base \
       libffi-dev
 # Install Radicale
-ADD https://github.com/Kozea/Radicale/archive/${VERSION}.tar.gz radicale.tar.gz
-RUN tar xzf radicale.tar.gz && \
+RUN wget --quiet https://github.com/Kozea/Radicale/archive/${VERSION}.tar.gz && \
```
Unrud: GitHub sends the ``Content-Disposition`` HTTP header with an alternative filename:
```
Content-Disposition: attachment; filename=Radicale-master.tar.gz
```
At the moment **wget** has only experimental support for this and it's disabled by default. Once it gets enabled however, the file will be saved as as ``Radicale-${VERSION}.tar.gz`` instead of ``${VERSION}.tar.gz`` and the command will fail.
You should specify the target filename explicitly with something like ``-O radicale.tar.gz``.

jaesivsm: Good catch, I wouldn't have saw that coming for a million bucks. I made the change.

Unrud:
Thank you.

PR #684, opened 2017-08-16
Skip comment lines in htpasswd files

mfherbst:
Some webservers like [nginx](https://nginx.org/en/docs/http/ngx_http_auth_basic_module.html#auth_basic_user_file) support comments in htpasswd files. Afaik Apache itself does not allow this.

In either way I think this would be a really useful feature for organising the accounts inside the htpasswd file a little, whilst it neither makes the code much more complicated nor any slower ;).

### DIFF
--- a/radicale/auth.py
+++ b/radicale/auth.py
@@ -211,6 +211,8 @@ class Auth(BaseAuth):
         try:
             with open(self.filename) as fd:
                 for line in fd:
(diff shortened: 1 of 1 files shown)


Unrud:
Fixed in f912642c20d91c5e5b9baea91981929f5715feae

PR #699, opened 2017-09-07
Add ldap auth

nebulade:
This tries to be mostly in line with the old ldap auth when it comes to the config files variables.

### DIFF
--- a/config
+++ b/config
@@ -70,7 +70,7 @@
 [auth]
 
 # Authentication method
-# Value: none | htpasswd | remote_user | http_x_remote_user
+# Value: none | htpasswd | remote_user | http_x_remote_user | ldap
 #type = none
 
 # Htpasswd filename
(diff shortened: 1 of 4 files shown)


nebulade:
This was started out as a plugin, but I would prefer that to be in the main code base, given that plugins simply tend to bitrot. I do however understand if the ldap3 module dependency or the additional code to maintain are not in line with the core project goals.

Also I haven't used python for years, so I expect several comments on the style and language use :-)

nebulade:
FYI we have now released the Cloudron package with this patch and this was also my main motivation for it: https://cloudron.io/store/org.radicale.cloudronapp2.html

Unrud:
I don't know anything about LDAP. There is already are simpler LDAP plugin: https://github.com/marcoh00/radicale-auth-ldap

> I do however understand if the ldap3 module dependency or the additional code to maintain are not in line with the core project goals.

I don't want to integrate plugins into Radicale.

I did take a quick look at your code and I found some problems:

## Code injection
```python
final_search_filter = self.ldap_filter.replace("%username", login)
```
This allows an attacker to inject an arbitrary string into the search query. You have to escape ``login``.

## Mapping login to user
You use **ldap_filter** in ``map_login_to_user`` to search for ``login`` and map it to ``user``.
Then you use **ldap_filter** in ``is_authenticated`` to search for ``user``. This obviously doesn't work, if ``login`` and ``user`` are not the same.
We could add the method ``is_authenticated2(login, user, password)`` to the authentication plugin interface.

## Raising exception when login doesn't exist
``map_login_to_user`` raises an exception when the search result is empty. This causes Radicale to answer with INTERNAL_SERVER_ERROR for non-existent users, instead of asking for another login.
First of all this is problematic because clients may not ask users for credentials again and it allows an attacker to probe for existing users.
With ``is_authenticated2`` you could just ``return login`` and check the user attribute again to prevent races.

nebulade:
Thanks for the answer and the feedback on the code, sadly I didn't find that plugin upfront since there was much noise about the old (version 1.x) ldap part. Would have been good to have a list of plugins/extensions somewhere, albeit as indicated personally I am not a fan of externally kept plugins after having to deal with so many related issues with them on other apps through the Cloudron project.

For the user mapping, the reason is, that in our case we allow the user to login with email or username, this is required in the is_authenticated() to get the unique identifier (in this case the DN)

In either way it seeems this can be closed as it was indicated that it wont be part of the core. I will see how to use the proposed plugin.

PR #732, opened 2017-10-23
Add notice about python3-setuptools to tutorial.md

Omniscimus:
While trying to install Radicale on a Debian server, I got an error message saying that the module `python-setuptools` was not installed. Running `sudo apt-get install python3-setuptools` fixed it. This might be helpful to other people who experience the same issue.


Unrud:
In Debian ``python3-setuptools`` only recommends ``python3-pipi``. That probably caused your problem.

I don't think that we should add it to the documentation. That's a problem of Debian packaging and the helpful error message makes it easy to fix.

Issue #522, opened 2016-10-02
Create a Docker registry in hub.docker.com


raulgd:
Hi,

I'd like to use Radicale on my server, and all my services are dockerized.

I see on this code repo that there's a docker file, it would be nice if you guys created a free public docker hub account with the radicale image so it's very easy to get Radicale up and running.

Also, in the meantime, if someone could share detailed instructions on buildind and running Radicale Docker would be very nice, as I'm seeing the docker file and I don't know which is the volume where the data will be saved, so I can map that path to the docker host hard drive.

Thanks

cshoredaniel:
I've created one based on the Dockerfile in Kozea/Radicale GitHub @ https://hub.docker.com/r/cshore/radicale/

funkyfuture:
fyi, here's one for the Raspberry Pi that also allows setting configuration parameters per environment variables: https://hub.docker.com/r/funkyfuture/rpi-radicale/

for an official image there could/should be variants with dependencies for the different storage and authentication backends.

amancevice:
Any update on this? It's pretty easy to set up an acct with Docker hub.

bebehei:
So, there is already [kozea](https://hub.docker.com/u/kozea/) registered with an automated build. So I guess at least one member of the kozea organisation has to have access to this dockerhub account.

In combination with my PR, I think this should get finally solved.

jasle:
I would be interested in a official docker image for radicale on the docker hub too.

Is there any help needed to get radicale on docker hub?

Trolldemorted:
Can we get an official response on this? Are there official docker images available, if yes where? Will you push (security) updates in a timely manner?

If not, did you give up or are there problems we can help with? I see multiple open, but ignored docker-related PRs.

dessalines:
This would be really nice to have.

sjal:
+1

instantdreams:
Please consider creating an official docker image and hosting on ghcr.io and hub.docker.com or other locations.

PR #756, opened 2017-12-13
Dockerfile

bebehei:
Like many others, I'd like to see a docker image from Kozea on docker hub (#522). I'm not talking about a docker library image (which often also get called official image).

On the first step, there are some neccessary technical changes. As dockerhub builds out of the git repository and handles git tags transparently in combination with docker image tags, the Dockerfile should build from the gitrepo and not download from github master again.

Also, I added an entrypoint, which executes radicale as a normal user and also sorted the package lists (as recommended in [dockerfile's best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/)).

---

After merging this, it should be a quite easy move to create an organization on dockerhub with an automated build repo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,4 +25,4 @@ VOLUME /etc/radicale
 # TCP port of Radicale (Publish it on a host interface!)
 EXPOSE 5232
 # Run Radicale (Configure it here or provide a "config" file!)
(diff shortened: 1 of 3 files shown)


bebehei:
I'm going to close this in favor of #768 There's no reason to favor this PR instead of @funkyfuture's work.

PR #766, opened 2018-01-09
Use built-in python UUID creation over custom-made implementation

djmattyg007:
This removes a custom-made implementation of UUID v4, which is one less
piece of code to worry about.

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -41,8 +41,8 @@ from hashlib import md5
 from importlib import import_module
 from itertools import chain, groupby
 from math import log
-from random import getrandbits
 from tempfile import NamedTemporaryFile, TemporaryDirectory
+from uuid import uuid4
 
 import vobject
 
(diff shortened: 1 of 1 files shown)


Unrud:
The reason for using a custom implementation is, that ``str(uuid.uuid4())`` is slow. With Python 3.3 it takes almost 10 times as long. Newer versions of Python are not as bad, but it's still significantly slower.

Sorry for taking a long time to answer this.

djmattyg007:
I would expect it to be slower. Python's `random` module does not generate cryptographically secure random numbers. However it feels like a microoptimisation. Both operations happen in the order of microseconds.

The real issue is that there's no guarantee that UUIDs generated by the home-grown version are actually universally unique, because they're based on a pseudo-random number generator. Python's built-in `uuid` package does provide this guarantee.

Unrud:
> However it feels like a microoptimisation.

Actually this has a very noticeable impact, when a big calendar/addressbook is uploaded. This was the reasons for replacing ``uuid.uuid4()`` with the primitive ``random_uuid4()``.

> Python's built-in ``uuid`` package does provide this guarantee.

It doesn't have to be cryptographically secure. It only has to be unique within the collection (we even check for collisions). The function name ``random_uuid4`` is a bit misleading, as it's not a "real" version 4 UUID.

djmattyg007:
If it's only unique with the collection, it's a globally unique ID.

I can't imagine there is such a large impact on modern computers. If you've seen it struggle when creating a lot of IDs all at once, it's likely that the available pool of entropy is drained fairly quickly, which would slow things down. That shouldn't be a reason to avoid using it though - it's just one of the side-effects of generating cryptographically secure random numbers.

If you're unwilling to merge this patch in its current state, perhaps we can reach a compromise. What if we have the code default to the secure RNG, but allow falling back to the insecure RNG by setting an environment variable?

PR #768, opened 2018-01-15
More Docker 

funkyfuture:
this PR sits on top of #756. @bebehei, i'd appreciate a review.

what still bothers me is the necessity to set `htpasswd` as authentication method. otherwise there'd be no need for any config file (apart from the users file) for a reasonable setup the covers. then entrypoint script could also write that config, given the file indcated by `RADICALE_CONFIG` doesn't exist. i'd say none authentication as a default is questionable anyway.

i'd also like to see more support for configuration via environment varables. (i've solved that once [here](https://github.com/funkyfuture/docker-rpi-radicale#interpolation)).

when this gets merged i will amend the docs.

sorry about the partly unrelated code changes, i can cherry pick these into another pr if it would help. reading commit by commit should be doable.


docker/Makefile
```diff
@@ -0,0 +1,12 @@
+.DEFAULT_GOAL := build
+
+IMAGE_NAME = kozea/radicale
```
bebehei: `:=`

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM python:alpine
 
 
 # metadata
-ENTRYPOINT ["/radicale/docker/entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini", "--", "/radicale/docker/entrypoint.sh"]
```
bebehei: As far as I understand, this is not necessary. All started processes are running in foreground, so the signals will get forwarded, too.

docker/Makefile
```diff
@@ -9,4 +9,4 @@ build:
 
 .PHONY: test-invokation
 test-invokation: build
-	docker run --rm $(IMAGE_NAME)
+	docker run --rm -i $(IMAGE_NAME) --debug
```
bebehei: This has nothing to do with this commit. Pls fixup the previous one.

bin/radicale
```diff
@@ -1,31 +0,0 @@
-#!/usr/bin/env python3
```
bebehei: Good change. But I guess, the chance to get this merged is higher in a separate PR.

docker/entrypoint.sh
```diff
@@ -3,9 +3,11 @@
...
+getent group  radicale &> /dev/null || addgroup -g ${RADICALE_GID} radicale
+getent passwd radicale &> /dev/null || adduser -D -H -s /bin/false -u ${RADICALE_GID} -G radicale radicale
 
-chown -R radicale:radicale "$(dirname "${RADICALE_CONFIG:-/etc/radicale/radicale.config}")" /var/lib/radicale
+RADICALE_CONFIG="${RADICALE_CONFIG:-/etc/radicale/radicale.config}"
 
-exec su - radicale -s /bin/sh -c "exec radicale --hosts 0.0.0.0:5232 ${RADICALE_CONFIG:+--config='${RADICALE_CONFIG}'} $*"
+chown -R radicale:radicale $(dirname $RADICALE_CONFIG) /var/lib/radicale
+
+exec su -s /bin/sh -c "exec radicale --hosts 0.0.0.0:5232 $*" radicale
```
bebehei: `RADICALE_CONFIG` env variable won't be visiable to the radicale process anymore. That's the reason I chose to pass it as argument `${RADICALE_CONFIG:+--config='${RADICALE_CONFIG}'}`

hooks/build
```diff
@@ -0,0 +1,15 @@
...
+
+# this script is used to build a Radicale Docker image locally
+# as well as on the Docker Hub.
+
+echo "------ HOOK START - BUILD -------"
+
+BASE_IMAGE=$(head -n1 Dockerfile | cut -d" " -f2)
+docker pull $BASE_IMAGE
+BUILD_DATE="$(date --rfc-3339 seconds)"
+VERSION=$(grep VERSION ../setup.py | cut -f3 -d" " | tr -d '"')
```
bebehei: `git describe --tags` might be better here, as it will also report an offset to the latest released version. So the patched version won't map to the same version number as the vanilla version.

Dockerfile
```diff
@@ -8,6 +8,18 @@ ENTRYPOINT ["/sbin/tini", "--", "/radicale/docker/entrypoint.sh"]
 EXPOSE 5232
 VOLUME ["/etc/radicale", "/var/lib/radicale"]
 
+ARG VERSION
```
bebehei: The `ARG` should be in a layer somewhere above the `apk` calls.

Dockerfile
```diff
@@ -30,10 +30,10 @@ RUN apk add --no-cache \
       build-base \
       libffi-dev \
       python3-dev \
- && python -m pip install \
-      bcrypt \
-      passlib \
- && apk del .builddeps
+ && python -m pip install passlib[bcrypt] \
```
bebehei: I'd still would have the package name split into a newline.

Dockerfile
```diff
@@ -23,16 +23,21 @@ LABEL org.label-schema.schema-version="1.0" \
...
+      apr-util \
       ca-certificates \
       openssl \
       tini \
  && apk add --no-cache --virtual .builddeps \
+      apache2-utils \
       build-base \
       libffi-dev \
       python3-dev \
+ && cp /usr/bin/htpasswd /tmp \
```
bebehei: Is there any reason, why htpasswd is needed? I'd advocate it to use it on the host itself.

hooks/build
```diff
@@ -10,6 +10,6 @@ docker pull $BASE_IMAGE
 BUILD_DATE="$(date --rfc-3339 seconds)"
 VERSION=$(grep VERSION ../setup.py | cut -f3 -d" " | tr -d '"')
 printenv
-docker build --build-arg BUILD_DATE="${BUILD_DATE}" --build-arg SOURCE_COMMIT="${GIT_SHA1}" --build-arg VERSION=$VERSION -t $IMAGE_NAME ..
+docker build --build-arg BUILD_DATE="${BUILD_DATE}" --build-arg SOURCE_COMMIT="${GIT_SHA1}" --build-arg VERSION=$VERSION --tag $IMAGE_NAME --file Dockerfile ..
```
bebehei: Setting defaults twice is always a bad idea.

docker/README.md
```diff
@@ -1,33 +1,49 @@
...
 - `/etc/radicale`: Radicale configuration (put the configuration file here)
 
-## Ports
+### Ports
+
+- `5323`: Radicale's http port
+
+### Creating a user in a running container
+
+    docker exec -ti <ontainer> htpasswd -B -C 16 /etc/radicale/users <username>
```
bebehei: You're running here as the root user and not the radicale user. You possibly have to restart the host or manually chown the file if it's fresh.

docker/compose/docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '2'
+
+services:
+  radicale:
+    image: kozea/radicale:2.1.8
+    ports:
+      - "5232:5232"
+    volumes:
+      - ./config:/etc/radicale
```
bebehei: You probably also want to mount the data?

bebehei:
Half of the review comments is outdated, because I've gone through your changes commitwise, but these changes still apply!

Also I had been nitpicking everywhere. Therefore some comments may not make sense, as the upcoming commits made logical changes, which invalidate my comments.

Also please remove from the commit messages the trailing periods. [This is a bad style](https://chris.beams.io/posts/git-commit/#end).

Dockerfile
funkyfuture: isn't it?

bebehei: Well, you set the `ARG`s with every build. So, the actual apk-calls (which are unparametrized), won't get cached at any time.

Dockerfile
funkyfuture: thanks for the hint. that's a leftover of several combinations i tested, also with `su-exec` which would be neat to use in the entrypoint script but unfortunately spawned an extra process in this context.

docker/Makefile
funkyfuture: what would be the effect?

docker/README.md
funkyfuture: d'oh. i had tested that against an existing file. i add the `--user` option to the example.

docker/compose/docker-compose.yml
funkyfuture: actually, no. for an evualuation setup that suffices imo.

docker/entrypoint.sh
funkyfuture: right, but i remove setting the default then as it sets an uncommon location. `/etc/radicale/config` is found anyway.

hooks/build
funkyfuture: which one do you mean?

funkyfuture:
> Also I had been nitpicking everywhere.

thanks for your helpful feedback. i'll ping back when i'm through with the rebase ~fuckup~ session.

Dockerfile
funkyfuture: the needed tool is not always avail. therefore it makes sense to distribute it with the image. it doesn't conflict with what you propose to advocate and i *guess* it's less than a megabyte space consumed.

funkyfuture:
the rebasing wasn't too bad actually. even figured out two tricks.

don't know about the prior commits. gh sometimes fucks stuff up. my rebase selection never reached that far.

hooks/build
bebehei: The default, you've actually set: `--file Dockerfile` This is the default and unneccessary.

docker/Makefile
bebehei: Well, it's a minor nitpick, like the `.DEFAULT_GOAL`.

funkyfuture: so, `=` and `:=` are semantically identical and your nitpicking on the inconsistence?

docker/Makefile
funkyfuture: for reviewers: this is fixed.

hooks/build
funkyfuture: no, i thougth that too, but in my environment it's necessary to use trhat option explicitly when `..` is used as context root.

docker/Makefile
bebehei: Yes.

bebehei: Yes.

hooks/build
bebehei: Oh, sh***. I haven't said anything 😉

hooks/build
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
+# this script is used to build a Radicale Docker image locally
+# as well as on the Docker Hub.
+
+echo "------ HOOK START - BUILD -------"
+
+BASE_IMAGE=$(head -n1 Dockerfile | cut -d" " -f2)
+docker pull $BASE_IMAGE
```
bebehei: Manual `docker pull` is not needed. `docker build` does it automatically. And if you need to update it all the time, you can use `docker build --pull`.

hooks/build
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
+# this script is used to build a Radicale Docker image locally
+# as well as on the Docker Hub.
```
bebehei: Dockerhub doesn't run uncontained shellscripts.

hooks/build
```diff
@@ -0,0 +1,15 @@
...
+
+echo "------ HOOK START - BUILD -------"
+
+BASE_IMAGE=$(head -n1 Dockerfile | cut -d" " -f2)
+docker pull $BASE_IMAGE
+BUILD_DATE="$(date --rfc-3339 seconds)"
+VERSION=$(git describe --tags)
+printenv
```
bebehei: A small note, worth to consider: `printenv | sort` might be better here.

hooks/build
funkyfuture: not sure what's your point here.

hooks/build
funkyfuture: yeah, that script was copied from another project where i build the image for one architecture.

hooks/build
bebehei: DockerHub just asks you, where your Dockerfile is located and directly builds the Docker image. You can't run this script on DockerHub. This comment is either just a minor mistake or - worse and I hope not - yields basic issues in your understanding of DockerHub, which will actually doom most concepts of this PR.

funkyfuture: nah, it dooms only the metadata as `label-schema.org` proposes. you're right, i had the whole time the misconception that how builds are performed [on the Docker Cloud](https://docs.docker.com/docker-cloud/builds/automated-build/#configure-automated-build-settings) is identical as on the Hub. however, i couldn't checkout whether this approach actually works when a build is configured on the Cloud frontend as i can't connect a repo there to a repo here (maybe one must pay for this).

so, i could just remove that metadata labels.

funkyfuture: oh, i just rechecked my local copies of an image that was [built](https://hub.docker.com/r/funkyfuture/deck-chores/builds/bhcoepocy5o7frwaswa8iee/) on the Docker Hub with the hooks a month ago, and it actually has the expected metadata. though the build log on the Hub look different.

from the discussion on [this issue](https://github.com/docker/hub-feedback/issues/508) i just realised that the hooks dir needs to be moved to the `docker` dir.

funkyfuture:
a'ight, i reworked this a little. essentially i reverted to an evaluation of the version for a docker build based on the contents of `setup.py`. my guess is that `git` is a noop oon the hub, unfortunately getting the version with the `describe` subcommand didn't work out..

i've recently build [deck-chores](https://github.com/funkyfuture/deck-chores) alike on the Docker Hub. and though [the logs](https://hub.docker.com/r/funkyfuture/deck-chores/builds/bfqfljxwqsdhm7pkpvvcxgk/) do not reflect the commands in the `build` file (but even the reported hash of the image matches). anyway, i can't explain why, but the metadata is added correctly. :-/

btw, the Docker stuff could also be in an extra repo. Radicale can also be installed from the cheesehop and there'd be no need to figure out the current version. btw, [here](https://github.com/moshez/msbdemo)'s an interesting approach to install Python apps in an image.

robertbeal:
Hmm noticed the commit for building the image from the latest commit. Is that a good idea? How will you tag built images? Would it not be better to build from official releases? Can understand that's harder to automate though.

There's a "production" grade image here I contribute towards:
https://github.com/tomsquest/docker-radicale

Might be worth combining in some way/form as it's all working towards the same goal, so that there's one true image instead of many.

docker/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+getent group  radicale &> /dev/null || addgroup -g ${RADICALE_GID} radicale
+getent passwd radicale &> /dev/null || adduser -D -H -s /bin/false -u ${RADICALE_GID} -G radicale radicale
+
+if [ -n "${RADICALE_CONFIG}" ]; then
+    chown -R radicale:radicale $(dirname $RADICALE_CONFIG) /var/lib/radicale
+    CONF_OPT="--config '${RADICALE_CONFIG}'"
+else
+    chown -R radicale:radicale /etc/radicale /var/lib/radicale
+    CONF_OPT=""
+fi
```
deronnax, line 15: why don't you do all that user stuff in the docker file itself?

docker/Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM python:alpine
```
deronnax, line 1: not pinning the python version?

deronnax:
no update on this?

docker/entrypoint.sh
funkyfuture: because the variable's values are to be configured per container.

funkyfuture:
i'm closing this as it seems to be not a project goal and is lost track here.

Issue #803, opened 2018-04-09
HTTP_X_SCRIPT_NAME not removed from PATH_INFO


drwilly:
I just migrated from radicale 1 to 2. My setup is nginx -> uwsgi -> radicale.
I used to rely on the `base_prefix` in radicale's config to get the `SCRIPT_NAME` and `PATH_INFO` parameters right which apparently is no longer possible.
My understanding of the parameter `HTTP_X_SCRIPT_NAME` is that it should overwrite `SCRIPT_NAME` and be stripped off of `PATH_INFO`. So if `HTTP_X_SCRIPT_NAME = /radicale` and `PATH_INFO = /radicale/myuser/` then the latter should be rewritten to `/myuser/`.
However that does not seem to be the case: https://github.com/Kozea/Radicale/blob/master/radicale/__init__.py#L404

Unrud:
This is the expected behaviour (see https://www.python.org/dev/peps/pep-0333/).
My suspicion is that **nginx** is not configured correctly: http://radicale.org/proxy/
The trailing slash causes **nginx** to strip ``/radicale`` from the path.

pbiering:
@chripo : is this still an issue with Radicale major version 3? If not, please close.

pbiering:
I've did a major review of the "apache" reverse proxy config example (supporting 4 different flavors) and stored a template in the repository now: https://github.com/Kozea/Radicale/blob/master/contrib/apache/radicale.conf

Can you please align your "nginx" configuration and check whether the issue disappears then?

PR #804, opened 2018-04-12
support HTTP_X_PATH_INFO

chripo:
there is an issue by detecting PATH_INFO like described in #803 
In Apache mod_proxy_fcgi module >=2.4.26 is a possibility to set / override PATH_INFO and SCRIPT_NAME per `ProxyFCGISetEnvIf` directive.
this patch allows to override PATH_INFO with HTTP_X_PATH_INFO for earlier versions and more restrictive setups.


pbiering:
@chripo : is this still an issue with Radicale major version 3? If not, please close.

pbiering:

PR #853, opened 2018-08-09
fix datetime parsing

aganzha:
Hello!

Just started Radicale in Docker and found that dates coming from DAVDroid are not parsed correctly.
Did not tested on other clients yet, so, perhaps it is not relevant to them.


Unrud:
Please create the pull request at https://github.com/eventable/vobject/

PR #862, opened 2018-08-24
Add missing cls paramter to class method.

tasn:
This fixes a bug introduced in 2.1.10. It was essentially causing Radicale to
break if the storage backend didn't not implement the newly added class
function.
This broke the EteSync backend.


tasn:
Seems to be fixed in master.

PR #875, opened 2018-09-23
[RFC][rights] Allow slash in sane_path

jspricke:
If the sane_path contains more then one slash, the rights backend
blocked the access, instead of granting rw permissions.

### DIFF
--- a/radicale/rights/authenticated.py
+++ b/radicale/rights/authenticated.py
@@ -30,6 +30,6 @@ class Rights(rights.BaseRights):
         sane_path = pathutils.strip_path(path)
         if "/" not in sane_path:
             return rights.intersect_permissions(permissions, "RW")
-        if sane_path.count("/") == 1:
+        if sane_path.count("/") >= 1:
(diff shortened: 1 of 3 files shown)


jspricke:
My storage plugin [radicale-remind](https://github.com/jspricke/radicale-remind) uses the real filesystem path in the URI. This could contain more then one slash and with 1bdc47bf it does not work anymore. Was this change on purpose or could we switch to something like `path.endswith("/")` maybe?
I can adopt the tests once we agreed on a solution.

Unrud:
The change will only be included in Radicale >=3, because it breaks backwards compatibility.
You can always use the ``from_file`` rights backend for custom rules.
I changed it, because clients can't automatically discover collections that are not direct children of the principal collection.

jspricke:
For me it's a regression from 2.1.9 where this was working and Davdroid discovers all descendants (didn't test 2.1.10 yet). So this sounds like a regression to me.

PR #880, opened 2018-10-08
Encryption for ics files

janochim:
Added support to encrypt the .ics files with AES256 in CFB mode. The resulting byte output from the encryption process is converted to a string so that the read function doesnt throw any errors. The IV is stored at the head of the encrypted file. The bcrypt encrypted password is sha256'ed to create a 32 bit long key for the encryption. I didnt implement a KDF because the keys are already stored in a safe way.

This option is only changeable before your first calendar entry, I havent implemented a conversion tool.
I expanded the programm because I have a couple of users in my server accessing the radicale server and I didnt want me to be able to read the calendar entries of other people.

This is my first contribution to an open source project, so I am not sure the code I wrote is considered 'good' :)

### DIFF
--- a/config
+++ b/config
@@ -106,6 +106,9 @@
 # Example: ([ -d .git ] || git init) && git add -A && (git diff --cached --quiet || git commit -m "Changes by "%(user)s)
 #hook =
 
+#Encrypt the ics files with AES256?
+#You can only specifiy this *before* you commit your first calendar entry
+encrypt = True
 
 [web]
 
--- a/radicale/config.py
+++ b/radicale/config.py
@@ -168,7 +168,11 @@ INITIAL_CONFIG = OrderedDict([
         ("hook", {
             "value": "",
             "help": "command that is run after changes to storage",
-            "type": str})])),
+            "type": str}),
+        ("encrypt", {
+            "value": "False",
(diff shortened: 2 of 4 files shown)


felixfontein:
A simple note: if you use the bcrypt'ed password to derive the key, anyone having access to the server who can read the file containing the hashes can also derive the key and decrypt the files. Thus this doesn't really seem to help in the scenario you sketched.

janochim:
mhmm, you're totally right...  I guess this is why no caldav server ever implemented it. You would need a seperate password from the authentication...  this was a totaly wasted effort then :/

felixfontein:
Hmm, in case the authentication is handled by Radicale itself (and not by a reverse proxy such as Apache or nginx), you might be able to get hold of the raw password. I'm not sure whether it is a good idea to include code which pushes the password around, though, since that increases the general attack surface.

janochim:
even if I did that, the problem you raised still persists, I could just add some code which reads the base64 encoded password from the http response and write  something that decrypts the files. I think it is impossible to make it impossible for me to access the files, since I am root on the server and can modify anything I want. so this pull request is useless... 
It was fun to play around with it though, learned alot :)

Issue #750, opened 2017-12-08
Can't import .ics file from OwnCloud


conchyliculture:
I installed Radicale in a Debian Sid environment with pip3
```
# radicale --version
2.1.8
```
I exported my calendar from owncloud as a  cal.ics file. I moved this file to my storage folder in Radicale:
`filesystem_folder/collection-root/user/cal.ics`, and made sure permissions were correctly sef.
After logging on the web interface, I am greeted with an error 500.

The Radicale log crashes with :

```
2017-12-08 11:10:25,266 - [7fbedb466700] INFO: Successful login: 'user'
2017-12-08 11:10:25,270 - [7fbedb466700] DEBUG: 'user' has read and write access to collection 'user'
2017-12-08 11:10:25,285 - [7fbedb466700] ERROR: An exception occurred during PROPFIND request on '/user/': Failed to load item 'cal.ics' in 'user': Item type 'VCALENDAR' not supported in generic collection
```
This shortened version of my cal.ics file also triggers the error:
```
BEGIN:VCALENDAR 
VERSION:2.0 
PRODID:ownCloud Calendar 0.6.4 
X-WR-CALNAME:cal 
BEGIN:VEVENT 
DTSTAMP:20140107T165713Z 
DTSTART;TZID=Europe/London:20140107T174000 
DTEND;TZID=Europe/London:20140107T184000 
SUMMARY:Lol 
DESCRIPTION:Test 
CLASS:PUBLIC 
STATUS:CONFIRMED 
UID:3cd9a5cb-1c0b-4311-b07b-39572de59d67-caldavsyncadapter 
BEGIN:VALARM 
TRIGGER;VALUE=DURATION:-PT120M 
DESCRIPTION:caldavsyncadapter standard description 
ACTION:DISPLAY 
END:VALARM 
BEGIN:VALARM 
TRIGGER;VALUE=DURATION:-PT60M 
DESCRIPTION:caldavsyncadapter standard description 
ACTION:DISPLAY 
END:VALARM 
END:VEVENT 
END:VCALENDAR 
```

Is there maybe another way to import ICS files in Radicale?

Unrud:
> I moved this file to my storage folder in Radicale

This doesn't work: [Storage Layout](http://radicale.org/storage/#layout)

> Is there maybe another way to import ICS files in Radicale?

```
curl -u 'user:password' -X PUT http://localhost:5232/user/calendar --data-binary @cal.ics
```

[...]

PR #892, opened 2018-11-14
Add manual import of existing calendar to documentation

GuiOm:
Hello,
I had to import an existing calendar, but I didn't find an how to in the documentation.
Based on your reply here : https://github.com/Kozea/Radicale/issues/750 I added the curl command in clients.md to import an ics file.

### DIFF
--- a/clients.md
+++ b/clients.md
@@ -159,3 +159,11 @@ Delete the collections by running something like:
 ```shell
 $ curl -u user -X DELETE 'http://localhost:5232/user/calendar'
 ```
+
+## Manual import of existing calendar
+
+To import an already existing calendar, download its ics file then :
+
+```shell
+curl -L -u "user" -X PUT http://localhost:5232/user/calendar --data-binary @calendar.ics
+```


Unrud:
The next major release will include an upload dialog in the web interface.

Biepa:
@Unrud How is the status, of the next major release.
Any possible dates?
Waiting for this feature :grimacing:

Tntdruid:
No updated for 7 months now

novoid:
Is my assumption right that this method does not _overwite_ the existing calendar content but it instead adding items to it?
If this holds true, then this is not a method where I periodically need to read-only sync radicale with a complete set of items of an external calendar, right?

Therefore my issue is (in other words): I do have a non-radicale calendar system like Org mode. I periodically export the whole calendar (not just updates) to an ics file. By using the method above, I assume that modified items do not get updated but duplicated instead. Correct?

Unrud:
> By using the method above, I assume that modified items do not get updated but duplicated instead. Correct?

The whole calendar gets overwritten.

PR #897, opened 2018-11-20
basic auto-configuration documentation.

birlorg:
add in some docs about how to setup auto-config.

The documented setup was tested against iOS and macOS.

### DIFF
--- a/clients.md
+++ b/clients.md
@@ -28,6 +28,53 @@ In some clients you can just enter the URL of the Radicale server
 enter the URL of the collection directly
 (e.g. `http://localhost:5232/user/calendar`).
 
+## Auto Configuration
+
+Some clients can "auto-configure" themselves, but it's not always supported by Radicale.  Here is what seems to work the best.
+
+Setup a hostname just for radicale (say dav.example.com) and have radicale answer at the web root (so SCRIPT_NAME would be empty in your proxy config if using a proxy.)
+
+setup cal and card dav DNS entries(in bind format):
+
+```
+_caldavs._tcp 86400 IN SRV 10 20 443 dav.example.com.
+_carddavs._tcp 86400 IN SRV 10 20 443 dav.example.com.
+```
+
+Which is saying, for service caldavs (secure caldav) talk on port 443 to host dav.example.com.
+and for carddavs (secure carddav) talk on port 443 to host dav.example.com 
+
+understanding of these values can be had here: https://en.wikipedia.org/wiki/SRV_record
+
+you have secure and insecure options for the service name, with the s at the end means secure mode (i.e. served over TLS/SSL, which you should always do!)
+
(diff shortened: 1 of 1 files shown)


Tntdruid:
Nice work

clients.md
```diff
@@ -28,6 +28,53 @@ In some clients you can just enter the URL of the Radicale server
...
+```
+#http://www.example.com/.well-known/caldav -> root of your dav server
+        location /.well-known/caldav/ {
+                return         301 https://dav.example.com$request_uri;
+        }
+        #http://www.example.com/.well-known/carddav -> root of your dav server
+        location /.well-known/carddav/ {
+                return         301 https://dav.example.com$request_uri;
+        }
+```
```
galaxy4public, line 72: The above `nginx` config does not match the comments around it.  I suggest to change the provided snippet to read as follows:
```
        #http://www.example.com/.well-known/caldav -> root of your dav server
        location /.well-known/caldav {
                return         301 https://dav.example.com/;
        }
        #http://www.example.com/.well-known/carddav -> root of your dav server
        location /.well-known/carddav {
                return         301 https://dav.example.com/;
        }
```

This way it matches the comments and should work as expected.

birlorg: Does it not work as shown, with $request_uri; ?

In my testing it does work.

birlorg:
I had forgotten I even wrote this.  It's been 2 years now, so I have no idea how well this has held up, it seems to still work for me, locally.  But I don't have any desire to fix the conflict, since there doesn't seem to be any desire for this to get merged upstream.

I'll leave it open, so it's easier for people to find if they are trying to do something like this for themselves.

clients.md
tieugene: This solution is a) nginx, 2) as proxy, 3) radicale in /
What about a) apache, b) wsgi application, c) virtual folder?

birlorg: @tieugene Feel free to write those docs.  I needed to solve this issue for myself, so I did, I happen to use nginx, as a proxy.  I wanted to help the project out as a thank you, and include what I did to make it work for me.

It's not been many years, so clearly there is either zero hurry or zero desire to have this moved into the official radical documentation.  So if you figure out how to make it work for apache/wsgi I still think it would be useful to include as a pull request.  Even if it's ignored, hopefully it will help other people struggling like yourself.

Me, I don't use Apache, so I have zero desire to figure it out for you.  Also I'm not remotely related to the project, just a user of the software.

pbiering:
@birlorg : are you able to rebase this PR to latest master?

birlorg:
@pbiering I don't use Radicale now, so I'm not sure if the docs still work as written, though I assume they do. I don't remember doing anything magical. You would probably be better off just making the changes as you see fit to the docs for yourself. I don't need or want credit or anything, feel free to steal as you see fit.

PR #910, opened 2018-12-13
Fix dockerfile with pip upgrade

jkuettner:
This commit fixes the docker build


Omniscimus:
👍 Let's merge 😄 

I had the same problem: because pip3 was outdated, the Docker build failed because it couldn't find the dependencies (in my case, cffi, which is a dependency for bcrypt). Updating pip3 solves the problem!

Unrud:
Thank you

Issue #929, opened 2019-02-25
Migration from 1.1.6 to 2.1.11 fails with single .isc calendar


leahoswald:
I use radicale for a some years now with single .ics files with my radical 1.1.1 on a debian stretch. I tried the export to be able to use version 2.1.11 (in buster). To export my old data I installed a radicale 1.1.6 in a virtual environment. The problem is, that the migration or export doesn't work, it fails with the following error and I could find nothing mentioning this error message.

```
INFO: Exporting storage for Radicale 2.0.0 to '/opt/radicale'
ERROR: Failed to find child collections of '/user/my-calendar.ics': generator raised StopIteration
```

audax:
I have the same problem. Any pointers?

audax:
I exported the calendar, made a new one with Radicale 2 and imported it again.
Not the most elegant solution, but good enough in my case, since I am the only user.

marcelvb:
@audax I have the same problem as @nazco. How did you export and import it exactly?

audax:
@marcelvb I exported the calendar from Thunderbird, deleted the CalDav sync from Thunderbird. Then I installed Radicale 2.x, without migrating any data, set up my account again, synced the empty calendar from the new Radicale install into Thunderbird and then imported the previously exported calendar into the newly synced calendar.

I had to fix some recurring events after that, but nothing major.

Tntdruid:
I always do it that way, works great.

marcelvb:
Thanks @audax. My problem was that I upgraded to Debian 10 (and Radicale 2) before making the export. So I could not access my calendar at all. I fixed it by using Docker to spin up an environment with Radicale 1, apply the above mentioned patch and export my data there. It took me hours to figure it all out, but it seems it all works now, finally.

ronnicek:
I have the same problem.. export ends with error:

```
INFO: Exporting storage for Radicale 2.0.0 to '/data/radicale/export/'
ERROR: Failed to find child collections of '/user/contacts.vcf': generator raised StopIteration
```

b-abadie:
I fixed it with 
```
diff --git a/radicale/__main__.py b/radicale/__main__.py
index 4c3c12a..e492de5 100644
--- a/radicale/__main__.py
+++ b/radicale/__main__.py
@@ -79,15 +79,16 @@ def export_storage(config, path, debug=False):
                     if debug:
                         traceback.print_exc()
                     continue
-                try:
-                    remaining_collections.extend(collection.children(
[...]

Issue #711, opened 2017-09-23
Mixture of collections and collection-root


Forage:
Hi,

With the release of Radicale 2 there's talk of `/var/lib/radicale/collections/` and `/var/lib/radicale/collection-root/`. The Radicale 1 export function exports to `collection-root` again.

It's not necessary causing an issue, but it does cause confusion and forces you to pay attention to not mix the two. It would be good to just stick to one and mention `collections` only as a side note if `collection-root` should be considered the new preferred location.

For some reason `/var/lib/radicale/collections/collection-root/` is also automagically created.
Accessing the web interface will add a nice error to the collection stating `[wsgi:error] [pid 24051] [7fa5935ce700] ERROR: An exception occurred during GET request on '/.web/': [Errno 13] Permission denied: '/var/lib/radicale/collections/collection-root'`

Forage:
Things are getting a bit more messy.

For the migration from version 1 to 2 I exported my collection and put the result in `/var/lib/radicale/collections/username/`.
Now my client can't find anything but Radicale decided to create `/var/lib/radicale/collections/collection-root/username/` with nothing in it.

I'm on Radicale 2.5 for the moment, and using WSGI with Apache. It is the latter asking me for the login credentials and I set the configuration option `type` in the `auth` section to `remote_user` for Radicale.
The web interface asks me for the same credentials again though. Using the web interface could also be the cause of the creation of `collection-root/username/` under `/var/lib/radicale/collections/`.

What should the proper structure be? Just `/var/lib/radicale/collections/username` or `/var/lib/radicale/collection-root/username`, no?

Unrud:
``/var/lib/radicale/collections`` is the default storage folder. It can be changed in the configuration.
The folder contains the lock file ``.Radicale.lock`` and the root collection in the folder ``collection-root``.

The root collection is not stored directly in the storage folder, but in the sub-folder ``collection-root``. This is different from Radicale 1.x.x. (It's needed for atomic creation, replacement and deletion of the root collection.)

> For the migration from version 1 to 2 I exported my collection and put the result in ``/var/lib/radicale/collections/username/``.

You have to keep folder ``collection-root``. The full path must be ``/var/lib/radicale/collections/collection-root/username/``.

Forage:
Thank you for your reply.

[...]

PR #970, opened 2019-07-18
Fix export on leaf collections

saimonn:
When exporting from 1.1.6 to 2.0.0, I had this error:
 ```
ERROR: Failed to find child collections of '/': generator raised StopIteration
```
Adding a check on is_node/is_leaf fixed that.

This may fix the same or similar export errors seen in  #711 and #929

### DIFF
--- a/radicale/__main__.py
+++ b/radicale/__main__.py
@@ -80,8 +80,9 @@ def export_storage(config, path, debug=False):
                         traceback.print_exc()
                     continue
                 try:
-                    remaining_collections.extend(collection.children(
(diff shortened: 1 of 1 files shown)


plroman:
Oh man, thanks a lot for the fix. It's very simple but I feel like you just saved me from a headache. I got the same error and that one test solved it.

I hope this PR gets merged.

copyrights:
Thanks @saimonn that works. 

Would be nice to have a 1.1.7 with this include for late migrating folk like me ;-)

PR #994, opened 2019-10-11
Add 'env' auth module

matthinc:
I wrote a small authentication module, which takes login and password from environment variables.

**Use-case**: This makes it really easy to setup a single-user-Radicale-instance in a dockerized environment.

Config:
```
[server]
# Bind all addresses
hosts = 0.0.0.0:5232

[auth]
type = env
```

A docker-compose.yml could look like this:

```
services:
  radicale:
    build: .
    ports:
      - 5232:5232
    volumes:
      - ./radicale:/data
      - ./radicale_config:/etc/radicale/
    environment:
      RADICALE_LOGIN: admin
      RADICALE_PASSWORD: test123
```

### DIFF
--- a/radicale/auth/__init__.py
+++ b/radicale/auth/__init__.py
@@ -58,7 +58,8 @@ from importlib import import_module
 
 from radicale.log import logger
 
-INTERNAL_TYPES = ("none", "remote_user", "http_x_remote_user", "htpasswd")
+INTERNAL_TYPES = ("none", "remote_user", "http_x_remote_user",
+                  "htpasswd", "env")
 
 
 def load(configuration):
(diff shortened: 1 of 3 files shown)


funkyfuture:
this would be indeed great for deployments with containers. i'm wondering whether multiple user should be definable by providing comma-separated values.

a `RADICALE_PASSWORD_<HASH_ALGORITHM>` as alternative to `…_PASSWORD` seems reasonable as well.

i could work on it, but there seems to be no interest.

pbiering (changes requested):
Please rebase to latest Radicale 3

pbiering:

PR #998, opened 2019-10-30
Remove configparser dependency when constructing Application object

ZipFile:
Loading auth, rights, storage and web inside of `__init__` may produce unexpected side effects. The fix is to move module loading logic to class method `from_config`. This change will make it possible to create application instances from third-party configs, useful when you want to embed Radicale into the existing framework.
Also, fixed instance variable naming to conform with PEP-8.


Unrud:
I don't see the point of this.

> Loading auth, rights, storage and web inside of ``__init__`` may produce unexpected side effects.

What side effects?

> This change will make it possible to create application instances from third-party configs

You can already do this. E.g.

```
# Load default configuration
config = configuration.load()
# Change the options you need
config.update({
    "logging": {
        "level": "debug"}}, "my custom config")
app = Application(config)
```

ZipFile:
>What side effects?

Importing something is a side effect on itself. It changes your `sys.modules` (at least). there might be spaghetti code hard to unwind if not imported in correct order, custom initializations, etc... But that's not the main point. Core idea is to avoid using file configs and `static_init`s. Config loader also assume certain module attribute to be present, it's really bothersome if you already have one's instance under the hand.

>You can already do this. E.g.

I've added an example to give you better understanding of my (and potentially others) use case. Please check. You can run it with `gunicorn --chdir examples ro_dummy_db:application`.

Unrud:
> Also, fixed instance variable naming to conform with PEP-8.

Good idea, I've applied it in a separate commit: 6202257fc290c90071bcd2eae1107ec7942545a3

> I've added an example to give you better understanding of my (and potentially others) use case.

Thank you, I understand now.

> Config loader also assume certain module attribute to be present, it's really bothersome if you already have one's instance under the hand.

I suggest a different approach. First we split the class ``BaseCollection`` into ``BaseStorage`` and ``BaseCollection``. To make the interface cleaner and especially get rid of the subclassing in the loader.
To address your specific problem we allow Callables for the configuration settings "auth.type", "rights.type"…, that return the instance the loader uses.

E.g.:
```python3
# Load default configuration
config = configuration.load()
# Change the options you need
config.update({
    "auth": {"type": Auth},
    "rights": {"type": lambda config: Rights(db)},
    …}, "my config")
app = Application(config)
```

ZipFile:
>First we split the class BaseCollection into BaseStorage and BaseCollection.

Makes sense.

>To address your specific problem we allow Callables for the configuration settings "auth.type", "rights.type"…, that return the instance the loader uses.

It feels like workaround. It does not solve initial problem: application class creating/loading its own dependencies.

Config/auth/rights/storage might be loaded and created using different subsystem (e.g. [ZCA](https://zopecomponent.readthedocs.io/en/latest/narr.html)). And it's just neat if you're not using one:
```
app = Application(
    auth=Auth(db),
    rights=Rights(),
    storage=Storage(db),
    web=Web(),
)
```
vs your approach.

ZipFile:
@Unrud I see you integrated some changes (pep8 + xmlutils encoding arg). Rebased.

ZipFile:
Removed config from builtin plugins.

ZipFile:
Rebased.

PR #1011, opened 2020-01-15
Allow callable in configuration for plugin.type

Unrud:
Example:

```python3
# Load default configuration
my_config = config.load()

# Pass a class directly
my_config.update({"auth": {"type": MyAuth}})

# Pass an object directly
my_rights = MyRights()
my_config.update({"rights": {"type": lambda config: my_rights}})

app = Application(my_config)
````

Closes #998

### DIFF
--- a/radicale/utils.py
+++ b/radicale/utils.py
@@ -23,6 +23,9 @@ from radicale.log import logger
 
 def load_plugin(internal_types, module_name, class_name, configuration):
     type_ = configuration.get(module_name, "type")
+    if callable(type_):
+        logger.info("%s type is %r", module_name, type_)
+        return type_(configuration)
     if type_ in internal_types:
         module = "radicale.%s.%s" % (module_name, type_)
     else:
--- a/radicale/config.py
+++ b/radicale/config.py
@@ -81,6 +81,12 @@ def list_of_ip_address(value):
     return [ip_address(s.strip()) for s in value.split(",")]
 
 
+def str_or_callable(value):
+    if callable(value):
+        return value
+    return str(value)
+
+
 def _convert_to_bool(value):
     if value.lower() not in RawConfigParser.BOOLEAN_STATES:
         raise ValueError("Not a boolean: %r" % value)
@@ -153,7 +159,7 @@ DEFAULT_CONFIG_SCHEMA = OrderedDict([
         ("type", {
             "value": "none",
             "help": "authentication method",
-            "type": str,
+            "type": str_or_callable,
             "internal": auth.INTERNAL_TYPES}),
         ("htpasswd_filename", {
             "value": "/etc/radicale/users",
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: radicale/utils.py [hunk 1]
```diff
@@ -23,6 +23,9 @@
 
 def load_plugin(internal_types, module_name, class_name, configuration):
     type_ = configuration.get(module_name, "type")
+    if callable(type_):
```
ZipFile: Plugins has clearly defined structure, plugins expect function interface consuming configparser's config and return an object implementing certain interface. When Radicale is used as standalone server it is perfectly okay for plugin loader to load plugins this way. But when you're not using plugin subsystem you should not enforce such behavior.

Unrud:
> What you're doing here is just ignoring config.

The configuration specifies the plugin to use. It's up to the plugin if it needs anything else from configuration.

> When you're integrating Radicale into the existing solution you already have a config provided by the framework and/or you do not want file configs at all

You don't need configuration files. You need an instance of ``config.Configuration``. Just look at it as a data class that is used instead of separate arguments. 

> Is it even possible for ``RawConfigParser`` to produce callable values?

``RawConfigParser`` is independent of ``config.Configuration``. If you call ``config.load()`` it's not used at all.
It's only included in ``config.load`` as a convenience if you want to use configuration files.

> Is it even possible for ``RawConfigParser`` to produce callable values?

No, you're not supposed to use it with configuration files.

ZipFile:
#998 clearly shows that (with minimal changes) you can use Radicale without `radicale.config`. Less code executed = less bugs, architecture improvements as a side bonus. `Why ignoring it if you can avoid using it at all?` question is still in power.

Unrud:
> ``Why ignoring it if you can avoid using it at all?`` question is still in power.

``config`` is a central place for validation of arguments. It's reused for command-line arguments, configuration files and ``config.Configuration``.
Plugins can provide their own config schemes for validation of their settings ([example](https://github.com/Kozea/Radicale/blob/8b7a680a1ec3b4b3a061bbec08ef41222f797472/radicale/tests/test_config.py#L144)).
We keep track of the configuration sources which provided a setting to give the user useful error messages (e.g. which setting from which configuration file is invalid).

If we don't use ``config.Configuration`` as a data class, we either have to duplicate all of the validation or don't check arguments at all.

ZipFile:
>`config` is a central place for validation of arguments. It's reused for command-line arguments, configuration files and config.Configuration.

That's only right if Radicale is used as standalone server. If you take a closer look at example #998 - there is no CLI either. CLI should be provided by framework, if needed. Neither Auth/Rights/Storage/Web are "plugins" too. In #998's example it up to *developer* to provide correct values.

>Plugins can provide their own config schemes for validation of their settings (example).
>We keep track of the configuration sources which provided a setting to give the user useful error messages (e.g. which setting from which configuration file is invalid).

Once again, in #998 it's a developer who is responsible for providing correct arguments. Not a user.

>don't check arguments at all.

This. There should be a documentation stating what values are expected. If developer does not follow documentation - it's his own fault.

I'm not against removing `radicale.config`. Telling users that they're doing something wrong is right. Existing config solution is perfectly okay for use as standalone server. But current application implementation makes it hard to impossible to avoid using `radicale.config` if you need nothing from it.

Issue #1017, opened 2020-02-15
Standalone FastCGI (also OpenBSD compatibility)


kousu:
I'm trying to run radicale on the [OpenBSD stack](https://bsd.plumbing/). I've decided FastCGI is the way to go, but I'm having trouble getting it to work well.

I've used radicale before in its default setup, running HTTP at https://myserver.com:5232/, but now I want to run it at https://myserver.com/radicale/ in parallel with other pieces of my site. I tried to follow [the nginx reverse proxy guide](https://radicale.org/proxy/), but eventually I realized that wasn't going to work because radicale needs the "/radicale" stripped, but HTTP proxies in relayd(8) don't (yet) support URL rewriting, and while httpd(8) does (`request strip 1`) it purposefully [*only* supports](http://man.openbsd.org/httpd.conf#socket) HTTP-to-FastCGI proxies. So I need to get Radicale running as a FastCGI.

Once I realized this I quickly found https://github.com/Kozea/Radicale/blob/master/radicale.fcgi. As is, it doesn't work with httpd(8) because

a. httpd(8) doesn't *spawn* FastCGI processes:
  * flup defaults to "a normal FastCGI" assuming that the HTTP server spawns the FastCGI and is in charge of replacing stdin with a socket. But OpenBSD encourages _standalone_ FastCGI processes which open their own sockets.

b. It does not daemonize itself
c. If it does daemonize itself, OpenBSD's `rcctl` infrastructure gets confused so that `rcctl stop radicale` and `rcctl check radicale` fail.
  * OpenBSD's daemon infrastructure works by `pgrep`ing for full command lines. It is expecting `/usr/local/bin/python3.7 /usr/local/bin/radicale` but with the shebang `#!/usr/bin/env python3`, radicale shows up like

    ```
    _radical 29507  0.0  0.2 20648 13672 ??  S       9:25PM    0:00.04 python3 /usr/local/bin/radicale (python3.7)
    ```
    
    whereas with `#!/usr/local/bin/python3.7` it shows up like
    
    ```
    _radical 87684  0.0  0.2 20520 13184 ??  S       9:26PM    0:00.03 /usr/local/bin/python3.7 /usr/local/bin/radicale
    ```

I ended up adapting it like this:

```
$ cat /usr/local/bin/radicale
#!/usr/local/bin/python3.7

from flup.server.fcgi import WSGIServer

import radicale
from radicale.__main__ import daemonize

if __name__ == "__main__":
[...]

PR #1018, opened 2020-02-17
WIP: Handle trailing slash root redirect problem

kousu:
See https://github.com/Kozea/Radicale/issues/1017#issuecomment-586928992

### DIFF
--- a/radicale/app/get.py
+++ b/radicale/app/get.py
@@ -19,7 +19,7 @@
 
 import posixpath
 from http import client
-from urllib.parse import quote
+from urllib.parse import quote, urlparse
 
 from radicale import httputils, pathutils, storage, xmlutils
 from radicale.log import logger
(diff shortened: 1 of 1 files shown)


kousu:
DOCUMENT_URI is openbsd-specific. Oops. I'll fix this up in a bit.

kousu:
Note to self: this is failing because (gunicorn? WSGI in general?) only promises `PATH_INFO`; no other FastCGI variables are reliable, apparently, e.g.:

```
{'PATH_INFO': '/',
 'REQUEST_METHOD': 'HEAD',
 'wsgi.errors': <_pytest.capture.EncodedFile object at 0x7fbefb8c5390>}
```

well I don't know what to do then. Maybe I can get OpenBSD to change how their httpd parses the "/path/to/app/" case, to distinguish it from "/path/to/app".

PR #1024, opened 2020-03-25
Fix group permissions

BartG95:
At least under Linux, the tmp file function generates a file that is only
readable by the current user, completely ignoring any previous permissions or
the umask.

That's annoying when you want to make backups. For example, I assign my backup-user to the radicale-group to make backups, assuming that files are group readable.

Previously I worked around this issue with 
````
hook = find . -type d -exec chmod g+rx {} \; ; find . -type f -exec chmod g+r {} \;
````
in my config. But that's a bit inefficient ;)

This patch uses any previous permissions if they exist.
Otherwise, it will make the file group readable if allowed by the umask.

I've tested this patch on my own installation and it works flawlessly.

### DIFF
--- a/radicale/storage/multifilesystem/__init__.py
+++ b/radicale/storage/multifilesystem/__init__.py
@@ -25,6 +25,7 @@ Uses one folder per collection and one file per collection entry.
 
 import contextlib
 import os
+import stat
 import time
 from itertools import chain
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: radicale/storage/multifilesystem/__init__.py [hunk 1]
```diff
@@ -83,6 +84,15 @@ def _atomic_write(self, path, mode="w", newline=None, sync_directory=True,
                 raise RuntimeError("Fsync'ing file %r failed: %s" %
                                    (path, e)) from e
             tmp.close()
+            if os.path.exists(path):
+                os.chmod(tmp.name, os.stat(path).st_mode)
```
Unrud, line 88: Multiple threads/processes can change the filesystem in parallel. The file might get deleted after the ``exists`` check.

radicale/storage/multifilesystem/__init__.py
```diff
@@ -83,6 +84,15 @@ def _atomic_write(self, path, mode="w", newline=None, sync_directory=True,
...
+            else:
+                tmp_mode = os.stat(tmp.name).st_mode
+                current_mask = os.umask(0)
+                os.umask(current_mask)
```
Unrud, line 92: Not thread-safe!

radicale/storage/multifilesystem/__init__.py
```diff
@@ -83,6 +84,15 @@ def _atomic_write(self, path, mode="w", newline=None, sync_directory=True,
...
+                if not (current_mask & stat.S_IRGRP):
+                    tmp_mode |= stat.S_IRGRP
+                    os.chmod(tmp.name, tmp_mode)
```
Unrud, line 95: The new mode should always be set to ``0o666 & ~current_mask`` to match the behavior of creating a file with ``open(...)``.

Issue #842, opened 2018-06-24
Calendar entry can not be moved - error 502


tiiiecherle:
Hey,

first of all thanks for the project.

While using on daily basis we encountered the following error when trying to move a calendar entry from one calendar to another.

```
„502“
at CalDAVMoveEntityQueueableOperation.
```

I use radicale on archlinux and the error occurs (reproducable) when trying to move an calendar event from one calendar to another (by double clicking the event and selecting another calendar) and both calendars belong to the same user. I use macos 10.13.5 calendar as a client. To reproduce create a new user in the radicale webinterface and add two or more calendars to the same user. Connect to the user from macos calendar, add a calendar entry to calendar1. Then double click it and select calendar2 of the same user as calendar.

Then an exclamation mark appears in the macos calendar and the above error message appears and asks for retrying or using the server version. Retrying doesn`t help either.

I think this is the corresponding entry from nginx log.

`172.16.2.32 - user [24/Jun/2018:11:26:24 +0200] "MOVE /radicale/user/7b82ff2b-efa1-f97f-99cd-c0e380876907/ACE78827-DC36-48E1-8FC1-8295D1D40E69.ics HTTP/1.1" 502 53 "-" "Mac+OS+X/10.13.5 (17F77) CalendarAgent/399.2.2"`

As a workaround you can cut the calendar event from calendar1,  select calendar2 in the sidebar of the macos calendar app and paste the entry to calendar2. This works without issues.

Would be very nice to see this fixed.

Thanks in advance.

tiiiecherle:
Actually this is a bit weird. I was looking for a solution for longer and now after posting the issue it seems I searched for the right things on google and with the help of the following "old" issues it seems solved.

[https://github.com/Kozea/Radicale/issues/774](https://github.com/Kozea/Radicale/issues/774)

[https://github.com/Kozea/Radicale/issues/838](https://github.com/Kozea/Radicale/issues/838)

My nginx block that seems to be working is now:

```
        location /radicale/ {
                proxy_pass https://127.0.0.1:5232/;
	        proxy_set_header X-Script-Name          /radicale;
	        #proxy_set_header X-Forwarded-For        $proxy_add_x_forwarded_for;
	        #proxy_pass_header                       Authorization;
                #set $destination $http_destination;
                #if ($destination ~* ^https(.+)$) {
                #        set $destination http$1;
                #}
                #proxy_set_header Destination $destination;
                proxy_set_header Host $host;
                # user authentication
[...]

PR #1037, opened 2020-05-03
Add $http_host Host header to nginx documentation

tgy:
This seems to cause issues with macOS Calendar, see
for example issue #842, where a 502 error is raised.

In my case, this was happening because of the
`HTTP_HOST` header being set to `localhost:PORT`
while my radicale was configured to be at some
`domain.tld` address.

I've tried it with the current `master` branch and
this seems to solve the issue with macOS Calendar not
being able to move the events from one calendar to
another.


Unrud:
Thank you

PR #1080, opened 2020-07-16
Fixing Dockerfile (pip3 and bcrypt)

namandixit:
The `Dockerfile` that ships with the project was broken, and did not install `pip3` or `bcrypt`. This PR installs `pip3` command using whatever package may be necessary (to prevent future breakage), and also installs `bcrypt` to allow secure authentication.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@ RUN apk add --no-cache \
       libffi-dev \
       ca-certificates \
       openssl
+# Install pip3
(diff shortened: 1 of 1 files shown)


Unrud:
Thank you, fixed in c6f6f2c1f4b1075aa95d26ea5012e5d1067ded39 and f3b6b4869d53ee368235732b3496a848baa0126e.

PR #1087, opened 2020-07-31
better dockerfile

deronnax:
* uses python-alpine rather than downloading python from APK repos
* uses a workdir (good practice)
* use local git checkout that is supposed to exist (that Dockerfile is in the git repo)
* install less deps
* merge install steps so it doesn't create separate big layers

Image size went from 329MB to 101MB, and on my machine, build time without cache went from 75 seconds to 45 minutes.

I was able to run the tests successfully from within the docker images and I was able to display the login page, so I think it works as good as before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,18 @@
-FROM alpine:latest
+FROM python:alpine
 
 # Version of Radicale (e.g. 3.0.x)
 ARG VERSION=master
 
+WORKDIR /radicale
+
+COPY . .
+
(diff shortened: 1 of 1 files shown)


deronnax:
some more good practices can be integrated:
* run as non-root (pretty important)
* copy even less files with a .dockerignore


But that's for another PR

deronnax:
I'm closing it as there is already a PR for a better docker image which goes further : #768

PR #1091, opened 2020-08-16
Adds config value to serve from subdirectory

oemel09:
This adds the possibility to serve radicale from a subdirectory, by specifying the subdirectory in the radicale config file.

I have the feeling, that the functionality to serve radicale from a subdirectory is implemented already, but somehow I couldn't find anything about it in the documentation and wasn't able to figure out how it is done.

### DIFF
--- a/radicale/app/__init__.py
+++ b/radicale/app/__init__.py
@@ -211,6 +211,8 @@ class Application(
             unsafe_base_prefix = environ["HTTP_X_SCRIPT_NAME"]
             logger.debug("Script name overwritten by client: %r",
                          unsafe_base_prefix)
+        elif self.configuration.get("server", "subdirectory"):
+            unsafe_base_prefix = self.configuration.get("server", "subdirectory")
         else:
             # SCRIPT_NAME is already removed from PATH_INFO, according to the
             # WSGI specification.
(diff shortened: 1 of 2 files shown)


Unrud:
> I have the feeling, that the functionality to serve radicale from a subdirectory is implemented already, but somehow I couldn't find anything about it in the documentation and wasn't able to figure out how it is done.

Yes, it depends on your setup.
If you are using a WSGI server, it should automatically set `SCRIPT_NAME` in the WSGI environ.
If you are using a reverse proxy, you can control it with the `X-Script-Name` header ([example in the documenation](https://radicale.org/3.0.html#tutorials/reverse-proxy)).

oemel09:
Thanks for your response, I saw the documentation and the reverse proxy.
My problem is, that I want to run radicale on a shared hoster where I don't have access to the Apache config, only the .htaccess files.
I used the following [tutorial ](https://lab.uberspace.de/guide_radicale.html) to install it, but I want to run it in a subfolder, not the root of my domain.
I couldn't find any method to add the `X-Script-Name` header.

Unrud:
> I couldn't find any method to add the `X-Script-Name` header.

Just add something like `RequestHeader set X-Script-Name /radicale/` to the .htaccess file.

Unrud:
I've added two examples for Apache .htaccess to the documentation:

* https://radicale.org/3.0.html#tutorials/reverse-proxy
* https://radicale.org/3.0.html#tutorials/reverse-proxy/manage-user-accounts-with-the-reverse-proxy

oemel09:
Thank you very much!

PR #1092, opened 2020-08-17
Hook capability for event changes and deletions

freakmaxi:
Implemented the none and RabbitMQ hook capability for events


Unrud:
I don't think this will work.

* Changes to the storage and running hooks are not atomic. Both can easily get out-of-sync in case of some error or system crash.
* The filesystem storage can get changed manually or by external programs. Hooks will never run for those changes.
* You are missing operations that change the storage (e.g. moving items)

My suggestion: Write a program that reads the filesystem storage folder directly and sends messages to RabbitMQ (as an example) for all new and deleted items since it was last run. Start the program via [storage->hook](https://radicale.org/3.0.html#documentation/configuration/storage/hook). Additionally run it on startup and at time intervals to catch external changes and crashes. (Don't forget to lock the storage.)

freakmaxi:
Honestly, manually or external program changes should be omitted because if we will consider touching files outside of the service, what will be the point of having the radicale. So I can write a program which will be the replacement of caldav service. Because of this reason, my hook implementation is just covering the changes that are triggered by the requests made to radicale.

Currently, I'm testing the service stability on my software. Around 30k people are using right now and planning to tune the performance for 5M people and also thinking to add clustering capability or piling/grouping by hash and make the service reachable by a reverse proxy head because one service will not be enough for all those people and radicale should be scalable somehow.

You are right, I've not implemented the hook for moving items.

PS: Nobody needed to approve this pull request. They can wait until I make it stabile after heavy usage or if they want, they can completely drop it. Returning the improvements on the code to the source is my responsibility. That's why we have this PR.

pbiering:
@freakmaxi : this changes a lot, is this still valid?

freakmaxi:
Hey,

Yes it is valid and working and help us to get the "change" notifications immediately in our backend services.

Thanks

pbiering:
> Yes it is valid and working and help us to get the "change" notifications immediately in our backend services.

In this case please rebase to Radicale 3 "latest"

pbiering:
No hurry here as being a new feature and if then integrated in 3.2.x

freakmaxi:
I told next week but plans are changed on my side. I've done it this week. it is currently ready. you can take a look

pbiering:
Is `pika>=1.3.2` a hard requirement? I'm asking because EPEL9 only supports currently 1.2.1 and Fedora 39 only 1.3.1, would mean in RPMs it must be bundled.

freakmaxi:
nope 1.1.0 and up is okay.

pbiering:
> nope 1.1.0 and up is okay.

can you please update your PR?

pbiering:
workflow check triggered but not successful, please investigate, details see above.

freakmaxi:
retry please

pbiering:
still not happy, next issue found...

freakmaxi:
it'll be okay this time. re-trigger please...

pbiering:
finally looks good, merged into new v3.2-devel branch

PR #1096, opened 2020-08-26
Fix permissions in storage folder (respect umask)

foresto:
Use temp file APIs that respect umask, so storage folder backups, git clones, etc. will work for members of the radicale group.

This is a backport of 630d49b7cfd04b0e3f963adb838b87563cf040d0
Related: #1024

### DIFF
--- a/radicale/storage.py
+++ b/radicale/storage.py
@@ -792,26 +792,18 @@ class Collection(BaseCollection):
 
     @contextmanager
     def _atomic_write(self, path, mode="w", newline=None, sync_directory=True):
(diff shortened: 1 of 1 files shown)


Unrud:
I consider this a minor problem and would rather not update the old version of Radicale. We no longer have CI testing for it.

pbiering:
@foresto : is this still a problem with Radicale major version 3?

pbiering:
> I haven't tested for it specifically, but a quick look at the collections managed by a Radicale 3.1 installation shows file permissions that look right.

In this case I will close this PR.

PR #1097, opened 2020-08-28
Web: add support for the POST HTTP method.

tasn:
This patch adds support for POST in addition to the already supported GET.
    
This is needed for implementing more complex web modules that also
support configuration modifications and advanced queries.


Unrud:
The current web plugin interface is very simple and easy to understand. Adding all `do_...` methods to it make it much more complex:

  * It would require a lot of documentation
  * It exposes the (currently) internal `do_...` methods to the public
  * Fallbacks must be implemented in the BaseWeb class to not raise `AttributeError`

It basically makes web plugins to WSGI applications themselves. In my opinion this can be addressed better at a higher level utilizing the WSGI server or implementing something like [WSGI Router](https://github.com/ericmoritz/wsgirouter).

> This is needed for implementing more complex web modules that also support configuration modifications

For simple use-cases URL variables can be used.

tasn:
> The current web plugin interface is very simple and easy to understand. Adding all `do_...` methods to it make it much more complex:
> 
>     * It would require a lot of documentation

How is it different to `get`? Isn't it exactly the same?

> 
>     * It exposes the (currently) internal `do_...` methods to the public

Aren't those just WSGI methods so therefore very documented?

> 
>     * Fallbacks must be implemented in the BaseWeb class to not raise `AttributeError`

I can add that if that's your concern, though really it should error if you are posting to `.web/` and the module doesn't implement it, and a server error seems appropriate to me.

> 
> 
> It basically makes web plugins to WSGI applications themselves. In my opinion this can be addressed better at a higher level utilizing the WSGI server or implementing something like [WSGI Router](https://github.com/ericmoritz/wsgirouter).

In a sense, yeah, it makes them into WSGI applications.
The same argument can be made to having the web module at all - it too can be better addressed at a higher level. The advantage of having this as a module is that you can implement useful web modules that people can just use (so call `radicale` with a configuration setting) rather than telling users they need to run and configure a UWSGI server.

> 
> > This is needed for implementing more complex web modules that also support configuration modifications
> 
> For simple use-cases URL variables can be used.

GET parameters are limited, but also GET requests shouldn't make persistent changes, so it's not standard to do so.

Unrud:
> How is it different to `get`? Isn't it exactly the same?

True.

> Aren't those just WSGI methods so therefore very documented?

They are not WSGI methods.

> GET parameters are limited, but also GET requests shouldn't make persistent changes, so it's not standard to do so.

Yeah, wouldn't adding the POST method be enough.

tasn:
> > Aren't those just WSGI methods so therefore very documented?
> 
> They are not WSGI methods.

Sorry, I misremembered.

> 
> > GET parameters are limited, but also GET requests shouldn't make persistent changes, so it's not standard to do so.
> 
> Yeah, wouldn't adding the POST method be enough.

Ha, so my original PR (before I pushed it) only had `post` implemented, but I assumed you'd want me to add the main methods too for completeness.

Would you like me to remove everything other than post, and add the default `post` implementation to the base web?

Btw, what I initially thought about doing is only call the web's `post` if it's implemented (`hasattr`), so if it doesn't exist the code just behaves the same way it does now. I'm happy to change it to that if you prefer.

tasn:
Hey, just following up on this.

Unrud:
> Would you like me to remove everything other than ``post``, and add the default post implementation to the base web?

Yes, and call the method in base web just ``post`` not ``do_POST`` to fall in line with the ``get`` method. I think it's best to return https://github.com/Kozea/Radicale/blob/9d25cc6c0ae6fa72bd6c57b76cbd25f1fddedbc0/radicale/httputils.py#L41-L43 for the default implementation.

You need to extend the tests in [radicale/tests/test_web.py](https://github.com/Kozea/Radicale/blob/master/radicale/tests/test_web.py) and [tests/custom/web.py](https://github.com/Kozea/Radicale/blob/master/radicale/tests/custom/web.py).

Maybe it would be useful to extract the methods [_read_raw_content](https://github.com/Kozea/Radicale/blob/9d25cc6c0ae6fa72bd6c57b76cbd25f1fddedbc0/radicale/app/__init__.py#L315) and [_read_content](https://github.com/Kozea/Radicale/blob/9d25cc6c0ae6fa72bd6c57b76cbd25f1fddedbc0/radicale/app/__init__.py#L324) to [radicale.httputils](https://github.com/Kozea/Radicale/blob/master/radicale/httputils.py). But this can also be done at a later point if the need arises.

tasn:
Adjusted according to your comments and pushed the changes.

The function was already called `post`, and not `do_POST`. The `do_POST` is the WSGI method name (same as for get), and thus can't be changed.

Please let me know if there's anything else.

radicale/app/post.py
```diff
@@ -0,0 +1,5 @@
+class ApplicationPostMixin:
```
Unrud, line 1: Please add the license/copyright notice to the file.

radicale/app/post.py
```diff
@@ -0,0 +1,5 @@
...
+    def do_POST(self, environ, base_prefix, path, user):
+        """Manage POST request."""
+        if path == "/.web" or path.startswith("/.web/"):
+            return self._web.post(environ, base_prefix, path, user)
```
Unrud, line 5: The method returns `None` for paths that do not belong to the web plugin. It should return METHOD_NOT_ALLOWED.

tasn:
Done.

Unrud:
Thank you.

tasn:
Great! Thanks! :tada:

PR #1102, opened 2020-09-10
ensure group 'radicale' gets created

### LINKED ISSUE #1101: Documentation: command to create user and group relies on system defaults
Running as a service => Linux with systemd system-wide

The command for creation of user `radicale` relies on system properties to get a user group with an identical name.
At least on OpenSUSE Tumbleweed the new user gets assigned to a different group by default.
We should add option `--user-group` to the example.

fe60:
Without the option --user-group the creation of the group depends on default values.
In OpenSUSE Tumbleweed the user 'radicale' became a member of 'users'.

This fixes #1101

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -213,7 +213,7 @@ requirements.
 ### Linux with systemd system-wide
 
 Create the **radicale** user and group for the Radicale service.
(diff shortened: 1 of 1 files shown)


Unrud:
Thank you.

PR #987, opened 2019-09-10
Update __init__.py

rradu92:
Changed log level from info to warning and add remote_host for authentication logs

Issue #918, opened 2019-01-10
How to correcty setup Log file?


Loader23:
Hi,
i want include Radicale log files in fail2ban but don't know how to set it up.
I have this log conf:
`[loggers]
keys = root

[handlers]
keys = file

[formatters]
keys = full

[logger_root]
# Change this to DEBUG or INFO for higher verbosity.
level = INFO
handlers = file

[handler_file]
class = handlers.RotatingFileHandler
args = ('/log/radicale.log', 'a', 100000, 3)
formatter = full

[formatter_full]
format = %(asctime)s - [%(thread)x] %(levelname)s: %(message)s
`

But the logfiles get spammed with this:
`2019-01-08 21:57:36,717 - [7fe02e170ae8] INFO: GET request for '/' received from 127.0.0.1 using 'curl/7.61.1'
2019-01-08 21:57:36,719 - [7fe02e170ae8] INFO: GET response status for '/' in 0.002 seconds: 302 Found
2019-01-08 21:58:06,846 - [7fe02e170ae8] INFO: GET request for '/' received from 127.0.0.1 using 'curl/7.61.1'
2019-01-08 21:58:06,849 - [7fe02e170ae8] INFO: GET response status for '/' in 0.003 seconds: 302 Found
2019-01-08 21:58:36,976 - [7fe02e170ae8] INFO: GET request for '/' received from 127.0.0.1 using 'curl/7.61.1'
2019-01-08 21:58:36,978 - [7fe02e170ae8] INFO: GET response status for '/' in 0.002 seconds: 302 Found
2019-01-08 21:59:07,104 - [7fe02e170ae8] INFO: GET request for '/' received from 127.0.0.1 using 'curl/7.61.1'
2019-01-08 21:59:07,106 - [7fe02e170ae8] INFO: GET response status for '/' in 0.002 seconds: 302 Found`

How do I set it up so that it only logs the relevant things?

kimpenhaus:
Hi,

just to be sure: in the main config file - in the `logging`section - do you have uncommented and filled the correct path to the `logging.conf` file?

For example like: 

```
[logging]

# Logging configuration file
# If no config is given, simple information is printed on the standard output
# For more information about the syntax of the configuration file, see:
# http://docs.python.org/library/logging.config.html
config = /config/logging.config
```

kimpenhaus:
Not sure what the `relevant` things are that you want to log, but have you tried configuring different level than `INFO`?

From the python docs - possible values are:

[...]

PR #1104, opened 2020-09-22
Failed login attempts enhanchment

AlphaJack:
Radicale now prints remote host IP address next to the failed login attempt, and uses [Warning] log level.

This makes the log easily parseable by fail2ban and allows to reduce the log verbosity.
Requested by #918 and #987


GreyMarvel:
This would great if it was merged in the 2.1xx branch also

PR #1131, opened 2021-01-02
some Dockerfile improvements

funkyfuture:
i know that Docker support is canny here, but i found that opportunity to reduce the images' size and couldn't resist to wipe some opinionated comments that don't fit all deployment environments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,7 @@ EXPOSE 5232
 # Run Radicale (Configure it here or provide a "config" file!)
 CMD ["radicale", "--hosts", "0.0.0.0:5232"]
 
-# Install dependencies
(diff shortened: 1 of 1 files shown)


felixfontein:
> wipe some opinionated comments that don't fit all deployment environments.

I'd rather understand them as instructions for beginners. People who know how Docker works will ignore them anyway.

funkyfuture:
> I'd rather understand them as instructions for beginners.

i do not think that these comments would help beginners at all. they might even endorse bad practices. proper instructions would be prose in the docs.

> People who know how Docker works will ignore them anyway.

i did not.

nullinger:
I did quite the same as @funkyfuture in my Dockerfile and follow his argumentation.

Unrud:
Thank you.

Issue #839, opened 2018-06-18
Non-ASCII passwords are not accepted on the web interface


vigiraud:
Non-ASCII passwords are not supported on the web interface, when authenticating. One get a «401 Unauthorized» even when entering the correct utf-8 password.

How to reproduce :
-  With **htpasswd -B -c /path/to/file user**, create a password file with an user in it. Bcrypt will be used. Put non-ASCII characters in the password, like «ç», «à», «œ» or «è» for example.
- Setup the rest of the radicale server, launch it, and go to its web page.
- Try to authenticate with the previously made account.

**htpasswd** seems to correctly handle utf-8 password, since verifying them with **htpasswd -v /path/to/file user** works. Also, clients like DAVdroid and Thunderbird actually works too. The issue probably resides in the web interface.

DietmarSchwertberger:
It's even worse: even ASCII characters are not supported when using `htpasswd_encryption = plain`.

The error message on the command line is
`ERROR: An exception occurred during GET request on '/': comparing strings with non-ASCII characters is not supported`

I am really surprised that the instructions on https://radicale.org/setup/ are more or less untested.

prcmt:
I get this error trying to use a users file of any kind - bycrypt or plain, even with purely ASCII characters.

PinkDev1:
2020-06-08. Succesfully replicated the issue with:
Radicale 3.0.0
Firefox 77.0.1 (64-bit)

How is this still an issue smh

dragtheron:
#1136 should resolve this issue.

Jeinzi:
I can't replicate this on an Arch server with:
- Python 3.11.3-2
- Radicale 3.1.8-2 or the current head d7ce2f0
- Firefox 115.0.2-1

I added a new user with the (bcrypt hashed) password "çàœè" and can use it to login to the web interface and create/delete collections. I've also been using KeepassXC "Extended ASCII" passwords since 2021-02-03 without a problem, but I'm not logging in to the webinterface that often.

pbiering:
if not reproducable, can this be closed now?

pbiering:
fixed now by https://github.com/Kozea/Radicale/pull/1136

PR #1136, opened 2021-01-08
Encode password to allow special characters

dragtheron:
XMLHttpRequest.open() does not automatically encode the password. Though it builds an basic auth schemed URI where '%' is the escaping indicator, thus passwords containing this characters are not accepted this way without manually replacing '%' with '%25' on the form.

### DIFF
--- a/radicale/web/internal_data/fn.js
+++ b/radicale/web/internal_data/fn.js
@@ -120,7 +120,7 @@ function Collection(href, type, displayname, description, color) {
  */
 function get_principal(user, password, callback) {
     let request = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


dragtheron:
Should resolve #839

Jeinzi:
Just tested this PR, works for me.
But as I commented in #839, I can't replicate the issue in the first place.

pbiering:
if not reproducable, can this be closed now?

pbiering:
Issue confirmed, funnywise partially also found on commerical platforms or devices, so take care by appending e.g. current `%YYYYMMDD` to a password - one should really avoid using "%XX" in all passwords.

Tested with `%20%33special`

Issue #1146, opened 2021-02-21
Bad request with recurrent events with EXDATE


symphorien:
**Steps to reproduce:**
with evolution:
Create an event with
* recurrence
* at least an exception (EXDATE)

For example
Here is how I graphically created the following ics file
[foo.zip](https://github.com/Kozea/Radicale/files/6017909/foo.zip)
![image](https://user-images.githubusercontent.com/12595971/108628737-65dd2a00-7454-11eb-89ef-d08ada8d6d71.png)
If I tell evolution to save it to a radicale server, then I get the following error:
```
Feb 21 15:46:44 blitiri radicale[31889]: [2021-02-21 15:46:44 +0100] [31889/Thread-552] [WARNING] Bad PUT request on '/symphorien/f9f54f38-827c-9f0d-c2b7-85f2312ad916/a9deac61a0d3ed4567a9e0c1fe9769c027b3f706.ics': can't compare offset-naive and offset-aware datetimes
```
Removing the EXDATE solves the problem.

----

Radicale 3.0.5 NixOS 20.09 Python 3.8

symphorien:
Radicale 3.0.5 NixOS 20.09 Python 3.8

Issue #957, opened 2019-06-02
Summary of problems and/or bugs creating and publishing a new calendar


jmfoleyjr:
This is a summary of the problems I reported in https://github.com/Kozea/Radicale/issues/956. Debug dumps, configs, etc. can be found there if anyone chooses to pursue these problem. 

I have a heterogeneous office environment with Windows 7, Window 10, Linux and Mac workstations, all members of an Active Directory domain. Thunderbird is a mail client on all the above types of workstations. In addition, some Windows 7 workstations use Outlook 2010 with CalDavSynchronizer. Some Windows 10 workstations use Outlook 365 with CalDavSynchronizer. Some Mac are running Outlook 365 (which currently has no non-Exchange calendar subscription capability) and the Macs also use Mac Calendar.

Here are my issues trying to create and publish to a new radicale calendar. Some of the problems may be usage issues on my part (and if so I would appreciated being straightened out on what I did wrong), and at least one I think is an actual bug: a  "400 Bad Request" error.

Problem 1: I attempted to create a new calendar from a user workstation. I could not do so via Thunderbird/Lightning. I tried specifying the URL of my desired new calendar: https://mydom.org:5232/OfficeCalendar. The radicale debug output is shown in the initial post of issue 956 (link listed above) and I'll not clutter this post by repeating that here. It seem to me that radicale should be able to create a new calendar from the calendar client program. If that is possible, and I did it wrong, please advise. If not possible I'd suggest that as a future enhancement as most calendar users are not technical.

Problem 2: Next I tried opening the URL https://mydom.org:5232 in a browser. I was not able to log in with the known correct credentials and got a "403 Access Forbidden" error (debug output in issue 956). The config uses auth type httpasswd and rights type from_file. This user is able to access published radicale calendars using these credentials, but not able to log into the URL with them. This may be a usage issue on my part, but if so I don't know what. The rights file is:
```
[owner-write]
user = .+
collection = %(login)s(/.*)?
permission = rw

[conference]
user = .+
collection = ConferenceCalendar(/.*)?
permission = rw

[office]
user = .+
collection = OfficeCalendar(/.*)?
permission = rw
```
Which was pretty much taken from https://radicale.org/rights/. If I disabled authentication and rights I was then able to log in and create the calendar.

[...]

Issue #956, opened 2019-06-01
How to publish existing WebDAV calendar to Radicale


jmfoleyjr:
This is surely a newbie question, but I've spent several hours searching and experimenting, so I need help. I have and existing WebDAV calendar .ics file. I want to now publish it to radicale/CalDAV. I have successfully created new CalDAV calendars with radicale, but they have been empty. I've tried simply publishing the WebDAV cal to https://mydom.org:5232/OfficeCalendar, but I get error (below). My rights file is:

```
# All authenticated users can update the Conference Room Calendar
[conference]
user = .+
collection = ConferenceCalendar(/.*)?
permission = rw

# All authenticated users can update the Office Calendar
[office]
user = .+
collection = OfficeCalendar(/.*)?
permission = rw

```
The ConferenceCalendar works fine, but that started empty and I didn't have any authentication or rights. The OfficeCalendar is the new one I want to create. When I tried publishing the WebDAV version of this, I used the URL https://mydom.org:5232/OfficeCalendar and got the following error with radicale -D:

```
[7f86cd85d700] DEBUG: Sanitized script name: ''
[7f86cd85d700] DEBUG: Sanitized path: '/OfficeCalendar'
[7f86cd85d700] DEBUG: Rule '':'OfficeCalendar' doesn't match '.+':'ConferenceCalendar(/.*)?' from section 'conference'
[7f86cd85d700] DEBUG: Rule '':'OfficeCalendar' doesn't match '.+':'OfficeCalendar(/.*)?' from section 'office'
[7f86cd85d700] INFO: Rights: '':'OfficeCalendar' doesn't match any section
[7f86cd85d700] DEBUG: Rule '':'' doesn't match '.+':'ConferenceCalendar(/.*)?' from section 'conference'
[7f86cd85d700] DEBUG: Rule '':'' doesn't match '.+':'OfficeCalendar(/.*)?' from section 'office'
[7f86cd85d700] INFO: Rights: '':'' doesn't match any section
[7f86cd85d700] INFO: Access to '/OfficeCalendar' denied for anonymous user
[7f86cd85d700] DEBUG: Asking client for authentication
[7f86cd85d700] DEBUG: Response content:
Access to the requested resource forbidden.
[7f86cd85d700] INFO: PUT response status for '/OfficeCalendar' in 0.002 seconds: 401 Unauthorized

```
How should I do this?

jmfoleyjr:
[...]

Issue #925, opened 2019-02-10
400 Bad Request when uploading a calendar (TypeError: can't compare offset-naive and offset-aware datetimes)


cpitclaudel:
Thanks for the hard work on Radicale!

I saw the closed issue at https://github.com/Kozea/Radicale/issues/806, but I'm running into this issue despite upgrading to the latest radicale (2.1.11). I exported my Gmail calendar to ICS, and when using curl to upload it to Radicale I see get a 'bad request' response.  The log says this:

```
Feb 10 17:46:34 raspi env[1326]: BEGIN:VEVENT
Feb 10 17:46:34 raspi env[1326]: CREATED:20100713T145922Z
Feb 10 17:46:34 raspi env[1326]: LAST-MODIFIED:20190210T151327Z
Feb 10 17:46:34 raspi env[1326]: DTSTAMP:20190210T151327Z
Feb 10 17:46:34 raspi env[1326]: UID:040000008200E00074C5B7101A82E00800000000000000002F6DF480DC3E534DBA44DA
Feb 10 17:46:34 raspi env[1326]:  50F42BCF512F6DF480DC3E534DBA44DA50F42BCF51
Feb 10 17:46:34 raspi env[1326]: SUMMARY:Maths
Feb 10 17:46:34 raspi env[1326]: STATUS:CONFIRMED
Feb 10 17:46:34 raspi env[1326]: DTSTART;TZID=America/New_York:20080203T040000
Feb 10 17:46:34 raspi env[1326]: DTEND;TZID=America/New_York:20080203T050000
Feb 10 17:46:34 raspi env[1326]: X-LIC-ERROR:No value for LOCATION property. Removing entire property:
Feb 10 17:46:34 raspi env[1326]: SEQUENCE:0
Feb 10 17:46:34 raspi env[1326]: TRANSP:OPAQUE
Feb 10 17:46:34 raspi env[1326]: END:VEVENT
Feb 10 17:46:34 raspi env[1326]: BEGIN:VEVENT
Feb 10 17:47:09 raspi env[1326]: [7601d470] WARNING: Bad PUT request on '/clement/4cf4ec72-8df4-8b1d-9608-ce8e09b30cc1/': Failed to store item '653243a0-5af3-d18e-5ca8-70f48e4c5f16.ics' in temporary collection 'clement/4cf4ec72-8df4-8b1d-9608-ce8e09b30cc1': can't compare offset-naive and offset-aware datetimes
[...]

PR #1159, opened 2021-04-17
Make "can't compare offset-naive and offset-aware datetimes" non-fatal

### LINKED ISSUE #758: Error when syncing calendar after migration to 2.1.8
I recently migrated from 1.x to 2.1.8, following the instructions on http://radicale.org/1to2/. However, now one of my calendars won't sync (several others do sync correctly).

The server-side error:
`ERROR: An exception occurred during PROPFIND request on '/alexander/jobb.ics/': Failed to load item 'anru8fl75lv98n2g1kiqvpfmps_google.com-20150326T162132Z.ics' in 'alexander/jobb.ics': Main component missing`

The client-side error:
https://gist.github.com/ezeenova/c3a39b8e5c2d80d5d8e5a82ca538b530

Any ideas what the problem might be?

chrysn:
This turns what used to result in a 500 error on loading calendars with
repetition rules that make vobject trigger comparison errors (typically
imported from old calendars, possibly created using faulty software)
into at worst ignoring the event when fetching a time slice; showing a
warning in the logs.

---

This showed up when I tried to migrate an almost 10-year-long calendar into radicale. Quite possibly the files are even faulty in the sense that time zones are missing where there should be some, but that shouldn't bring down the whole calendar (as it so far did).

Might help close #758 and #956 / #957 (as they seem to be the ones suffering from this error cropping up in imports); others (#925, #1146) may profit from a generalization of this.

Note that there may *also* be an underlying bug in the vobject library, but that shouldn't stand in the way of fixing this here where it can catch all subtle variations of it. That can happen in parallel; if vobject catches all of these cases, no warnings will be shown any more out of radicale.

### DIFF
--- a/radicale/item/filter.py
+++ b/radicale/item/filter.py
@@ -189,6 +189,21 @@ def visit_time_ranges(vobject_item, child_name, range_fn, infinity_fn):
 
     """
 
+    def wrap_getrruleset(iterator):
+        """The vobject.icalendar.RecurringComponent.getrruleset method
+        occasionally compares timestamps in an unchecked fashion, leading to
+        TypeError("can't compare offset-naive and offset-aware datetimes")
+        errors that break the whole REPORT query when raised unchecked.
+
+        This iterator wraps such iterators, and terminates them safely rather
+        than raising the error; a warning is shown instead."""
+        try:
+            yield from iterator
+        except TypeError as e:
+            logger.warning("Recurrence calculation ran into comparison error "
+                    "(%r), ignoring iteration data.", e)
+            return
+
     # HACK: According to rfc5545-3.8.4.4 an recurrance that is resheduled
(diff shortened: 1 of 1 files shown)


chrysn:
Looking through unrelated other import errors, for a more stringent handling it may be an option to add a check to `check_and_sanitize_items` to look for whether all is timezoned or not as it should be.

Given that the vobject repetition iterator may still be faulty somewhere or we may miss a check, I still think that silencing them here is a practical thing; the more that's done in the checking stage, the fewer warnings will show, and maybe they can be turned to escalate when they haven't come up for a time.

chrysn:
I've found it easier to just fix the offending files -- in all cases, the `EXDATE;VALUE=date:20120101` lines were missing a `Z` at the end, when all the other dates had proper time zones attached.

Issue #1158, opened 2021-04-10
documentation: log to file


ohreally:
Hi,

The documentation on the website for version 3 doesn't document how to log to a file, and neither does the configuration file.

Cheers,
  Rob

pbiering:
PR need rebase

pbiering:
related PR already closed as no response received.

PR #1165, opened 2021-05-21
Added FileHandler for logging

ekaats:
Solves #1158

Added a FileHandler for logging. Two things I am not sure about:

1. Not sure if custom config entry is used properly
2. Used the default FileHandler as I am not sure if I should have used some kind of threaded handler.

### DIFF
--- a/config
+++ b/config
@@ -110,6 +110,8 @@
 # Value: debug | info | warning | error | critical
 #level = warning
 
+logfile = /var/log/radicale.log
+
 # Don't include passwords in logs
 #mask_passwords = True
 
--- a/radicale/config.py
+++ b/radicale/config.py
@@ -218,6 +218,11 @@ DEFAULT_CONFIG_SCHEMA = OrderedDict([
             "value": "warning",
             "help": "threshold for the logger",
             "type": logging_level}),
+        ("logfile", {
+            "value": "/tmp/radicale.log",
+            "help": "logfile to write to",
+            "type": filepath}
+        ),
         ("mask_passwords", {
             "value": "True",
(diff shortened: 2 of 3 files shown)


embeddedc:
@Unrud Any reason why this shouldn't be merged?

radicale/config.py
```diff
@@ -218,6 +218,11 @@ def _convert_to_bool(value):
             "value": "warning",
             "help": "threshold for the logger",
             "type": logging_level}),
+        ("logfile", {
+            "value": "/tmp/radicale.log",
```
ehem, line 222: I am **not** part of the Radicale team, so this is a quick review by a random developer.

The string "/tmp/radicale.log" is considered True in Python.  As such this will by default enable logging to "/tmp/radicale.log".  As "/tmp" is a world-writable directory, if any other users exist on the system they can cause Radicale to overwrite any file writable by Radicale.  In other words, I'm pretty sure this is a security hole.

config
```diff
@@ -110,6 +110,8 @@
 # Value: debug | info | warning | error | critical
 #level = warning
 
+logfile = /var/log/radicale.log
```
pbiering, line 113: this can trigger SELinux issues, better would be /var/log/radicale/radicale.log + a proper logwatch configuration for rotation, in case no autorotation is implemented

radicale/config.py
pbiering: Logging to `/tmp` in an example is ugly.

pbiering:
@ekaats : please rebase to latest Radicale 3 and also check all comments.

pbiering:

Issue #718, opened 2017-10-03
Authentication Methods Unclear


dwhagar:
So, is the _only_ authentication method that is actually usable, `htpasswd`?

The config file that comes with the distribution clearly states:

    [auth]
    # Authentication method
    # Value: None | htpasswd | IMAP | LDAP | PAM | courier | http

Of course, setting the `type=` to IMAP, LDAP, or PAM shows that there are no Python files for those authentication methods.  Taking a look at the actual code, it clearly shows that such a module has yet to be written.  From what I read on another issue, this has been intentionally done but the config file suggests it has not?  What gives?  Why was this removed?  Is there a workaround?

Tntdruid:
Can only use htpasswd, the OS dist is. outdated version

http://radicale.org/setup/

kousu:
Hiya,

I'm also wondering what the plan here is. Would you be open to accepting a new Radicale 3.0-compatible `auth/pam.py` and `auth/imap.py` files? As recently as half a year before this thread (https://github.com/Kozea/Radicale/issues/552#issuecomment-281982826) PAM was still supported.

pbiering:
@kousu : if required, please submit PR for authentication using PAM or IMAP against latest 3.x

PR #1175, opened 2021-08-08
Support OpenBSD authentication

kousu:
This assumes that on OpenBSD you will grant privileges to Radicale with `usermod -G auth _radicale`. Presumably when Radicale 3.0 is finally packaged for OpenBSD that will be handled in the package and users won't have to think about it.

For now, to use this, boot/install OpenBSD, and do:

```
$ doas pkg_add py3-pip
$ doas pip install -U git+https://github.com/kousu/Radicale/@bsdauth
$ doas useradd -m -d /var/lib/radicale _radicale && doas usermod -G auth _radicale
$ doas -u _radicale python3 -m radicale --auth-type=bsdauth --storage-filesystem-folder=/var/lib/radicale/collections
```

then visit http://localhost:5232 *and login in as the same user you use for ssh*.

This is part of addressing https://github.com/Kozea/Radicale/issues/718

### DIFF
--- a/radicale/auth/__init__.py
+++ b/radicale/auth/__init__.py
@@ -30,7 +30,7 @@ Take a look at the class ``BaseAuth`` if you want to implement your own.
 
 from radicale import utils
 
-INTERNAL_TYPES = ("none", "remote_user", "http_x_remote_user", "htpasswd")
+INTERNAL_TYPES = ("none", "remote_user", "http_x_remote_user", "htpasswd", "bsdauth")
 
 
 def load(configuration):
--- a/radicale/auth/bsdauth.py
+++ b/radicale/auth/bsdauth.py
@@ -1,0 +1,71 @@
+# This file is part of Radicale Server - Calendar Server
+# Copyright © 2008 Nicolas Kandel
+# Copyright © 2008 Pascal Halter
+# Copyright © 2008-2017 Guillaume Ayoub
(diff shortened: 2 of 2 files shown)


kousu:
I don't know if this should be trying to check `uname` or something else to be able to warn users if they have accidentally configured `bsdauth` on a non-bsdauth system.

bukhalo:
@kousu personally thx for `bsdauth` implementation! If I understand correctly, Radicale has kicked out all third-party authentication methods since version 2. See: https://radicale.org/2.1.html#documentation/migration-from-1xx-to-2xx/authentication Therefore, I can assume that the author will never accept these changes. 

Wouldn't it be better to put these changes in the authentication plugin? As written here https://radicale.org/3.0.html#documentation/plugins and some examples from PyPi https://pypi.org/search/?q=radicale&page=1

kousu:
Thanks @bukhalo. I'm withdrawing this and I'll maintain it as a plugin, instead, like you suggest.

kousu:
I've ported this to a [plugin](https://pypi.org/project/radicale-bsdauth/)! @bukhalo (or anyone else), if you have some time and happen to be in the same niche I am, I'd appreciate some feedback on it :). Sorry it took me 6 months to get around to it...

Install it:

```
doas pip install radicale-bsdauth
``` 

and add this to your `/etc/radicale/config`:

```
[auth]
type = radicale_bsdauth
```

PR #1187, opened 2021-10-14
Add py.typed to mark as having typing information

jspricke:
According to PEP 561.


Unrud:
Adding the file to **MANIFEST.in** only includes it into the source archive but it doesn't get installed.

It should be added to `package_data`:

https://github.com/Kozea/Radicale/blob/35e7ee5a08671eeb764e070acecee6c25356e81e/setup.py#L77

Unrud:
Thank you

PR #1207, opened 2022-01-11
config & rights: use open() for better error messages

lotheac:
ConfigParser().read() doesn't differentiate between different types of
failure to read files, causing eg. "No such file" to be logged in all
cases, for example if permissions are insufficient. fix that by using
open() and ConfigParser().read_file() instead.

### DIFF
--- a/radicale/config.py
+++ b/radicale/config.py
@@ -276,17 +276,18 @@ def load(paths: Optional[Iterable[Tuple[str, bool]]] = None
     for path, ignore_if_missing in paths:
         parser = RawConfigParser()
         config_source = "config file %r" % path
+        config: types.CONFIG
         try:
-            if not parser.read(path):
-                config = Configuration.SOURCE_MISSING
-                if not ignore_if_missing:
-                    raise RuntimeError("No such file: %r" % path)
-            else:
+            with open(path, "r") as f:
+                parser.read_file(f)
                 config = {s: {o: parser[s][o] for o in parser.options(s)}
                           for s in parser.sections()}
         except Exception as e:
-            raise RuntimeError("Failed to load %s: %s" % (config_source, e)
-                               ) from e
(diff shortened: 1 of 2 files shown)


radicale/config.py
```diff
@@ -277,13 +277,15 @@ def load(paths: Optional[Iterable[Tuple[str, bool]]] = None
...
-                config = Configuration.SOURCE_MISSING
-                if not ignore_if_missing:
-                    raise RuntimeError("No such file: %r" % path)
-            else:
+            with open(path, 'r') as f:
+                parser.read_file(f)
                 config = {s: {o: parser[s][o] for o in parser.options(s)}
-                          for s in parser.sections()}
+                        for s in parser.sections()}
+        except FileNotFoundError as e:
```
Unrud, line 284: I propose something like the following to avoid the duplicate:
```python
try:
    …
except Exception as e:
    if not (ignore_if_mssing and isinstance(e, FileNotFoundError)):
        raise RuntimeError(…)
    config = Configuration.SOURCE_MISSING
```

radicale/config.py
```diff
@@ -277,13 +277,15 @@ def load(paths: Optional[Iterable[Tuple[str, bool]]] = None
         parser = RawConfigParser()
         config_source = "config file %r" % path
         try:
-            if not parser.read(path):
-                config = Configuration.SOURCE_MISSING
-                if not ignore_if_missing:
-                    raise RuntimeError("No such file: %r" % path)
-            else:
+            with open(path, 'r') as f:
```
Unrud, line 280: We use `"` instead of `'`.

radicale/config.py
lotheac: thanks, fixed.

radicale/config.py
lotheac: thanks, fixed.

Unrud:
Thank you.

Issue #1144, opened 2021-02-01
Filtering address-book properties does not work


muellerj:
With large address books (images inside `*.vcf`) the overhead of sending each contact in full is significant. As such, if the client only requests some of the fields, we should only send those. As an example, the following request query should trigger a response only containing the fields `FN`, `EMAIL` and `UID`:

```
<CR:addressbook-query xmlns="DAV:" xmlns:CR="urn:ietf:params:xml:ns:carddav">
  <prop>
    <CR:address-data>
      <CR:prop name="FN" />
      <CR:prop name="EMAIL" />
      <CR:prop name="UID" />
    </CR:address-data>
    <getlastmodified />
    <getetag />
  </prop>
  <CR:filter>
    <CR:prop-filter name="FN" />
  </CR:filter>
</CR:addressbook-query>
```

If you are interested, I can take a stab at an implementation.

muellerj:
... and if anyone could give me a pointer on where to start, I'd appreciate it very much.

muellerj:
Note to self, this is the relevant piece of info from the logs

```
[2021-01-24 12:46:46 +0000] [26164/Thread-3] [DEBUG] Request content:
<?xml version="1.0"?>
<CR:addressbook-query xmlns="DAV:" xmlns:CR="urn:ietf:params:xml:ns:carddav">
  <prop>
    <CR:address-data>
      <CR:prop name="FN" />
      <CR:prop name="EMAIL" />
      <CR:prop name="UID" />
    </CR:address-data>
    <getlastmodified />
    <getetag />
  </prop>
  <CR:filter>
    <CR:prop-filter name="FN" />
  </CR:filter>
</CR:addressbook-query>

[2021-01-24 12:47:05 +0000] [26164/Thread-3] [DEBUG] Response content:
<?xml version="1.0"?>
<multistatus xmlns="DAV:" xmlns:CR="urn:ietf:params:xml:ns:carddav">
  <response>
    <href>/jonas%40tigger.cloud/36c4a029-32b5-b4ec-259a-d11e9d76c049/36cd24fa-7358-42d8-8891-bc36952bfeda.vcf</href>
    <propstat>
      <prop>
        <CR:address-data>BEGIN:VCARD
        [...]
```

pbiering:
assumed working now with the provided commits.

PR #1221, opened 2022-02-23
Allow filtering address-book properties in report

muellerj:
See Issue #1144


muellerj:
I just did some more testing on this - reducing the number of fields does not significantly reduce the response time, probably because it's dominated by I/O of reading the cards and processing them.

A significant improvement would have to have the cards in cache/in a database.

Closing this.

PR #1250, opened 2022-07-02
Fix docs: default value of `htpasswd_filename`

schnerring:
Fix mismatch between default values of `htpasswd_filename` in `config.py` and `DOCUMENTATION.md`:

https://github.com/Kozea/Radicale/blob/c14defcba87b7fb83e0556f1000706cdd21b230e/radicale/config.py#L165-L168


Unrud:
Thank you

PR #1262, opened 2022-09-26
Dockerfile improvements

eugene-davis:
* runs as non-root user
* allows specifying which "extras" dependencies to install (defaults to bcrypt)
* adds ENTRYPOINT (splitting off of CMD)
* adds a dev-Dockerfile to install from the current commit rather than from a release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,12 @@ FROM python:3-alpine as builder
 # Version of Radicale (e.g. v3)
 ARG VERSION=master
 
+# Optional dependencies (e.g. bcrypt)
+ARG DEPENDENCIES=bcrypt
+
 RUN apk add --no-cache --virtual gcc libffi-dev musl-dev \
     && python -m venv /app/venv \
-    && /app/venv/bin/pip install --no-cache-dir "Radicale[bcrypt] @ https://github.com/Kozea/Radicale/archive/${VERSION}.tar.gz"
+    && /app/venv/bin/pip install --no-cache-dir "Radicale[${DEPENDENCIES}] @ https://github.com/Kozea/Radicale/archive/${VERSION}.tar.gz"
 
 
 FROM python:3-alpine
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,30 @@
 # This file is intended to be used apart from the containing source code tree.
 
-FROM python:3-alpine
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,17 +1,33 @@
...
 # Run Radicale
-CMD ["radicale", "--hosts", "0.0.0.0:5232"]
+ENTRYPOINT [ "/app/bin/python", "/app/bin/radicale"]
+CMD ["--hosts", "0.0.0.0:5232"]
 
-RUN apk add --no-cache ca-certificates openssl \
- && apk add --no-cache --virtual .build-deps gcc libffi-dev musl-dev \
- && pip install --no-cache-dir "Radicale[bcrypt] @ https://github.com/Kozea/Radicale/archive/${VERSION}.tar.gz" \
- && apk del .build-deps
+USER 1000
```
h3xx, line 33: Wouldn't it be better to use `USER radicale`?

Dockerfile
```diff
@@ -1,17 +1,33 @@
...
+
+
+FROM python:3-alpine
+
+WORKDIR /app
+
+RUN adduser radicale --home /var/lib/radicale --system --uid 1000 --disabled-password \
+    && apk add --no-cache ca-certificates openssl
+
+COPY --chown=1000 --from=builder /app/venv /app
```
h3xx, line 23: Wouldn't it be better to say `COPY --chown=radicale [...]`?

h3xx (changes requested):

Dockerfile
eugene-davis: Indeed, updated in both places in both Dockerfiles

funkyfuture:
for the sake of simplicity, i meanwhile use this:

```Dockerfile
FROM debian

EXPOSE 5232
CMD ["radicale", "--hosts", "0.0.0.0:5232"]

RUN apt update \
  && apt install --yes --no-install-recommends python3-bcrypt python3-tz radicale \
  && rm --force --recursive /var/lib/apt/lists/*

COPY config users /etc/radicale/
```

the idea of a container (venv) in a (linux) container just isn't compelling.

yes, it wouldn't hurt to let another user run the process.

h3xx (approved):

gentooboontoo:
Improvements rebased and merged into `master` branch.

Issue #1217, opened 2022-02-14
macOS / iOS Calendar.app client support?


seanm:
More of a question than a bug report, but... the list of supported clients here:

https://radicale.org/v3.html#supported-clients

does not include the macOS / iOS Calendar applications.  Is this an oversight, or are they indeed not supported?

Thanks.

alqu:
Hm, I am running radicale ONLY with MacOS and iOS clients. Works flawlessly.

Though there is one caveat: you have to import the SSL-Certificate of your Radicale-Server manually to your iOS / Mac beforehand to make it work (only if you created the certificate by yourself).

I am running Radicale behind an NginX-proxy.

seanm:
Thanks for that info!  So I guess this is a bug report afterall, against the documentation.

I note that the [baikal](https://sabre.io/baikal/) docs say macOS is supported... It almost caused me to discard radicale...

fredthomsen:
@alqu I have been running radicale v2 for sometime and tried to get macOS working with v2 and a new instance of v3 I spun up. I've spun this up on my local network as a test, no proxying or SSL is enabled.  Neither seem to work rather I just get "Unable to verify account name or password".  Logs seem to indicate the server is not getting hit with every request to login which I am confused by.  Is there some simple trick I am missing?

alqu:
Yes, there is a trick. To make it actually work on your own server with your self signed certs, you need to manually import your CA-cert into the keystore. And then trust this cert.

Result: All certs that are created under that CA are automatically trusted - and then automagically the login works.

I did this with the CA's certs since IOS and/or MacOS trusts all certs that are issued by this CA. :-)

alqu:
...and you wrote "no SSL" - this could be also the culprit since this is a security measure by IOS to NOT trust unencrypted connections. Yeah. I took me a while to find that out.

fredthomsen:
Ah ok. Now I see a reference to this here as well: https://github.com/Kozea/Radicale/issues/967#issuecomment-511547885 explicitly mentioning macOS.  Thanks for the info.

lingdocs:
> Hm, I am running radicale ONLY with MacOS and iOS clients. Works flawlessly.
> 
> Though there is one caveat: you have to import the SSL-Certificate of your Radicale-Server manually to your iOS / Mac beforehand to make it work (only if you created the certificate by yourself).
> 
> I am running Radicale behind an NginX-proxy.

@alqu Will this work with a Letsencrypt certificate?

[...]

PR #1271, opened 2022-11-20
MOVE: consider HTTP_X_FORWARDED_HOST also supported in case Radicale is running behind reverse proxy

pbiering:
While `thunderbird` uses DELETE -> PUT sequence for moving a calendar item from one calendar to the other, iOS uses MOVE, which causing an issue in case of Radicale is running behind a reverse proxy.

Upstream version < 2022-11-20:

```
[1128/Thread-10331] [INFO] MOVE request for '/USER1/GROUP1.ics/UUID.ics' received from 127.0.0.1 (forwarded for 'CLIENT-IP') using 'iOS/16.1.1 (20B101) dataaccessd/1.0'
```

Move is rejected with "502"

```
[1128/Thread-10331] [INFO] Unsupported destination address: 'https://SERVER-FQDN/USER1/USER1.ics/UUID.ics'
[1128/Thread-10331] [INFO] MOVE response status for '/USER1/GROUP1.ics/UUID.ics' in 0.004 seconds: 502 Bad Gateway
```

This PR compares the host part of the given destination against the `X-Forwarded-Host` header if provided.

Once applied, MOVE is successful:

```
MOVE response status for '/USER1/USER1.ics/UUID.ics' in 0.373 seconds: 201 Created
```

### DIFF
--- a/radicale/app/move.py
+++ b/radicale/app/move.py
@@ -33,10 +33,16 @@ class ApplicationPartMove(ApplicationBase):
         """Manage MOVE request."""
         raw_dest = environ.get("HTTP_DESTINATION", "")
         to_url = urlparse(raw_dest)
-        if to_url.netloc != environ["HTTP_HOST"]:
-            logger.info("Unsupported destination address: %r", raw_dest)
-            # Remote destination server, not supported
-            return httputils.REMOTE_DESTINATION
+        if environ.get("HTTP_X_FORWARDED_HOST"):
+            if to_url.netloc != environ["HTTP_X_FORWARDED_HOST"]:
+                logger.info("Unsupported destination address: %r", raw_dest)
+                # Remote destination server, not supported
+                return httputils.REMOTE_DESTINATION
+        else:
+            if to_url.netloc != environ["HTTP_HOST"]:
+                logger.info("Unsupported destination address: %r", raw_dest)
+                # Remote destination server, not supported
+                return httputils.REMOTE_DESTINATION
         access = Access(self._rights, user, path)
         if not access.check("w"):
             return httputils.NOT_ALLOWED


pbiering:
btw: the test results shown above are not caused by the commit, looks like a general issue in the test setup

GaiusVampus:
Your patch seems to work as intended; at least it solves my issue reported in https://github.com/Kozea/Radicale/issues/1217#issuecomment-1164451975

Thank you very much for supplying this solution!

So from my point of view Radicale with your patch applied is fully compatible with iOS/macOS devices. We use  OpenBSD as a host and a manual installation of Radicale (not the official port which is still going with version 2 of Radicale) and an Apache proxy for the https handling that seems to be mandatory for today's Apple clients.

kbruegge:
Hey there. I’m no expert on this code base but this commit looks good as far as I can tell. It would really help with running behind a reverse proxy. Is there anything one could do to move forward with this?

pbiering:
@Unrud : are you sure that your implementation of `get_server_netloc` will proper work in case the reverse proxy in front of `radicale` is listen on non-standard port like e.g. 8443?

Unrud:
Non-standard ports should be included in the hostname.

pbiering:
I think this needs to be reopened, as issue were found in https://github.com/Kozea/Radicale/issues/1301

pbiering:
@Unrud at least for Apache reverse proxy following is required:

https://github.com/Kozea/Radicale/issues/1301#issuecomment-1528809515

```
        <If "%{HTTPS} =~ /on/">
                RequestHeader set X-Forwarded-Proto "https"
        </If>
```

But how to handle non-standard port on reverse proxy side (e.g. 8443)? The radicale daemon will not get noted about in the headers between Apache and Radicale by default.

-> a candidate would be requiring in this case an additional header like

```
   RequestHeader set X-Forwarded-Port "%{SERVER_PORT}s"
```

and extend code in move.py by (untested):

```
@@ -32,6 +32,7 @@
         host = environ["HTTP_X_FORWARDED_HOST"]
         proto = environ.get("HTTP_X_FORWARDED_PROTO") or "http"
         port = "443" if proto == "https" else "80"
+        port = environ["HTTP_X_FORWARDED_PORT"] or port
     else:
         host = environ.get("HTTP_HOST") or environ["SERVER_NAME"]
         proto = environ["wsgi.url_scheme"]
```

pbiering:
additional PR submitted: https://github.com/Kozea/Radicale/pull/1303

PR #1237, opened 2022-04-20
Drop timestamp and PID from log format.

kousu:
Radicale is designed to be run under a service manager of some kind, primarily systemd but also OpenBSD's rcctl, docker, kubernetes, etc. All of these log timestamps and some ID for the sending process.

Even on non-systemd systems, it's easy route the logs to syslog, with timestamp and PID attached using

    radicale 2>&1 | logger -ip daemon.info -t radicale

see: https://github.com/Kozea/Radicale/issues/1085/#issuecomment-1104532326

Basically, this is redundant:

```
[kousu@radio ~]$ sudo journalctl -f -u radicale
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] Loaded config file '/etc/radicale/config'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] Skipped missing config file '/.config/radicale/config'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] Starting Radicale
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] auth type is 'radicale.auth.none'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] storage type is 'radicale.storage.multifilesystem'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] rights type is 'radicale.rights.owner_only'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] web type is 'radicale.web.internal'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] Listening on '[127.0.0.1]:5232'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] Listening on '[::1]:5232'
Apr 20 16:48:46 radio.kousu.ca radicale[17569]: [2022-04-20 16:48:46 -0400] [17569] [INFO] Radicale server ready
```

so this makes the output instead:

```
[...]

PR #1276, opened 2023-01-11
in case started by systemd, log without timestamp

pbiering:
this PR disables logging of timestamp twice if started by systemd

this somehow covers at least one issue mentioned in https://github.com/Kozea/Radicale/pull/1237

### DIFF
--- a/radicale/log.py
+++ b/radicale/log.py
@@ -35,6 +35,7 @@ from radicale import types
 
 LOGGER_NAME: str = "radicale"
 LOGGER_FORMAT: str = "[%(asctime)s] [%(ident)s] [%(levelname)s] %(message)s"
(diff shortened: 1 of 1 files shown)


Unrud:
I think the method for detecting systemd is not specific enough.
* The environment variable `INVOCATION_ID` can also be set when Radicale is run from a third-party WSGI server that was started by systemd but does logging itself.
* Logging for the systemd unit could be set up to use a different method than forwarding to journald

The [`JOURNAL_STREAM`](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#$JOURNAL_STREAM) environment variable seems interesting. We can check if `wsgi.errors` stream is actually connected to systemd journal.

pbiering:
@Unrud : no problem if there is a improved/better proposal, but there was also a non-systemd related case which also need to be catched: https://github.com/Kozea/Radicale/pull/1237

Issue #1125, opened 2020-12-17
filtering for Categories property throws exception


lkopyt:
Hi,

i installed Radicale 3.0.6 to setup a calendar for maintenance events in my work environment.
querying the server for 'maintenance' events with prop-filter returns exception like this:

`[2020-12-16 17:00:27 +0100] [28940/Thread-27] [ERROR] An exception occurred during REPORT request on '/user1/438b326a-ef91-09d9-3ff4-e7d6c794b12c/': Failed to filter item 'f33a3f80dfbbe46ce8b4ae54a1fc1ce95ba5f47b.ics' from 'user1/438b326a-ef91-09d9-3ff4-e7d6c794b12c': 'list' object has no attribute 'lower'`

Categories property is defined in iCalendar RFC as comma-separated text.
Radicale seems to treat it as a list ? any bug or I am supposed to prepare a filter in some other way than text-match as below?

query:
```
curl -u user1 -X REPORT 'http://10.88.0.65:5232/user1/438b326a-ef91-09d9-3ff4-e7d6c794b12c/' --data \
'<?xml version="1.0" encoding="utf-8" ?>
   <C:calendar-query xmlns:D="DAV:"
                 xmlns:C="urn:ietf:params:xml:ns:caldav">
     <D:prop>
       <D:getetag/>
       <C:calendar-data/>
     </D:prop>
     <C:filter>
       <C:comp-filter name="VCALENDAR">
         <C:comp-filter name="VEVENT">
           <C:prop-filter name="CATEGORIES">
            <C:text-match collation="i;ascii-casemap">maintenance</C:text-match>
           </C:prop-filter>
         </C:comp-filter>
       </C:comp-filter>
     </C:filter>
   </C:calendar-query>'
```

icalendar event stored in Radicale:
```
# pwd
/var/lib/radicale/collections/collection-root/user1/438b326a-ef91-09d9-3ff4-e7d6c794b12c
# cat f33a3f80dfbbe46ce8b4ae54a1fc1ce95ba5f47b.ics
BEGIN:VCALENDAR
VERSION:2.0
CALSCALE:GREGORIAN
PRODID:-//Ximian//NONSGML Evolution Calendar//EN
BEGIN:VEVENT
UID:f33a3f80dfbbe46ce8b4ae54a1fc1ce95ba5f47b
DTSTART;VALUE=DATE:20201216
DTEND;VALUE=DATE:20201217
CATEGORIES:maintenance
CLASS:PUBLIC
[...]

PR #1277, opened 2023-01-20
Handle lists in filter values for CATEGORIES

lbt:
Issue #1125

Signed-off-by: David Greaves <david@dgreaves.com>


pbiering (commented):
looks good

PR #1289, opened 2023-03-20
Upgrade to journald's native journal protocol

Unrud:
Related to #1276

### DIFF
--- a/radicale/log.py
+++ b/radicale/log.py
@@ -25,11 +25,16 @@ Log messages are sent to the first available target of:
 
 """
 
+import contextlib
 import logging
 import os
+import socket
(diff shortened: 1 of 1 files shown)


pbiering:
found that this particular extension requires an SELinux policy extension in case of enabled

`allow radicale_t kernel_t:unix_dgram_socket sendto;`

would be helpful if the OSerror exception will log the message as found here after inserting some debug code:

`PermissionError: [Errno 13] Permission denied`

Unrud:
Good catch, fixed in 5998bae60e4d1c8adfcd9600e28af46c7a628e6b

PR #1290, opened 2023-03-21
User-selectable logger formats

Unrud:
Format can be set with the environment variables `RADICALE_LOG_FORMAT`. Available formats are `verbose` (default) and `journal`. If journald's native protocol is used, `journal` format is the default.

Closes #1276

### DIFF
--- a/radicale/log.py
+++ b/radicale/log.py
@@ -65,15 +65,17 @@ class IdentLogRecordFactory:
 
     def __call__(self, *args: Any, **kwargs: Any) -> logging.LogRecord:
         record = self._upstream_factory(*args, **kwargs)
-        ident = "%d" % os.getpid()
-        main_thread = threading.main_thread()
-        current_thread = threading.current_thread()
-        if current_thread.name and main_thread != current_thread:
-            ident += "/%s" % current_thread.name
+        ident = ("%d" % record.process if record.process is not None
+                 else record.processName or "unknown")
+        tid = None
+        if record.thread is not None:
+            if record.thread != threading.main_thread().ident:
+                ident += "/%s" % (record.threadName or "unknown")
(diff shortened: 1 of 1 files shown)


pbiering (approved):
confirmed working

kousu:
I'm wondering if there's a reason this was done as an environment variable instead of a config file option? Most of radicale is configured via its config file. The only envs Radicale uses are:

1. RADICALE_CONFIG, to define where its config file is
2. JOURNAL_STREAM, part of [_try_emit_journal()](https://github.com/Kozea/Radicale/blob/9f1e243f00151e1f8b68772a7fd9ba95a485b959/radicale/log.py#L140)
3. the one added here

<details>

```
[kousu@server Radicale]$ git grep os.environ
radicale/__init__.py:                              os.environ.get("RADICALE_CONFIG"))
radicale/__main__.py:            os.environ.get("RADICALE_CONFIG"),
radicale/log.py:            dev, inode = os.environ.get("JOURNAL_STREAM", "").split(":", 1)
radicale/log.py:    format_name = os.environ.get("RADICALE_LOG_FORMAT") or None
radicale/server.py:    os_environ: MutableMapping[str, str] = {}
radicale/tests/test_server.py:            env={**os.environ, "PYTHONPATH": os.pathsep.join(sys.path)})
radicale/tests/test_server.py:        env = os.environ.copy()
```

</details>

I know, 12 factor-app is a nice standard in theory and all, but setting per-daemon envs is actually kind of [tricky on OpenBSD](https://marc.info/?l=openbsd-misc&m=168205958524158&w=2). It's maybe a bit better on systemd, I'll grant, since you can write an /etc/systemd/system/override.d/radicale.conf that just has one line to set the env.

Do you think anyone would be opposed to me (or someone at least) patching to also look for this option in the config file?

Issue #662, opened 2017-07-19
Expand Recurring Events


balrok:
Hello, I want to write a very simple client, for which I need the expanded date of a recurring event as in https://tools.ietf.org/html/rfc4791#section-7.8.3

This was also once asked in 
https://stackoverflow.com/questions/37711699/expanding-recurring-events-in-caldav

When retrieving the events of next week and having the recurring date expanded already by the server, it would simplify a lot.

Best Regards
Carl

bgervan:
Hi, is this feature planned in the near future? Can we help in anything?

Forger125:
Happy to help provide any additional info, I am running into this problem from the HomeAssistant CalDAV integration when using my home Radicale server, only displaying the first event in a series.

pbiering:
- Is there any reference implementation as source code available for supporting this feature for "standard" reoccurring schedules + also related single shifted/canceled/adjusted ones?

- Also one has to consider a limit (and a related error message) to avoid unexpected denial-of-server attack. Think about a daily event but requested time range is e.g. 2 years -> resulting in 730 returned entries...this can even overflood a client implementation...haven't read through the RFC, but there should be 2 limits somehow:
  - max entries
  - max size (think about a schedule which has additional information stored inside, because this will be replicated in the mentioned example 730 times)

tvogel:
This is the expand code from Sabre that is used by NextCloud:

https://github.com/sabre-io/vobject/blob/49dbd1bacba784025cc012fd7d908dd87a725914/lib/Component/VCalendar.php#L280

I donot see any arbitrary "safety" limits in that. Neither in the code leading up to that method, starting here:

https://github.com/sabre-io/dav/blob/264f567943ad904d26709fa31bcc3dcc263f3a89/lib/CalDAV/ICSExportPlugin.php#L83

Also, I would be quite hesitant because it violates data consistency. If you request daily events for two years, then it is correct to report 730 items with full detail, so any precautions should be optional and used with quite some care as the client needs to be made aware how to still request all data (by shortening the start/end interval and iterating.

pbiering:
> Hi, is this feature planned in the near future? Can we help in anything?

based on reference code from above, one has to implement that extension into radicale, I cannot do that at the moment, did on first shot not even found where to insert...but I can test a Pull Request.

pbiering:
Support is now implemented by https://github.com/Kozea/Radicale/pull/1292, please try master branch and file comment on PR if still a problem

PR #1293, opened 2023-03-30
Processing expand property

metallerok:
https://github.com/Kozea/Radicale/issues/662


pbiering:
@metallerok : can you please explain more details about this PR

metallerok:
In accordance with https://www.rfc-editor.org/rfc/rfc4791#section-9.6.5 (9.6.5 CALDAV:expand XML Element), a REPORT request can specify the `expand` property, which forces the server to expand recurring components into individual recurrence instances for a specified period. 
Example https://www.rfc-editor.org/rfc/rfc4791#section-7.8.3

pbiering:
@metallerok : would it be able to create also a test case for this kind of support?

pbiering:
@metallerok : thank you for the PR, but looks like the test run is not happy...please check and update your PR.
See also https://github.com/Kozea/Radicale/wiki/Contributions

pbiering (approved):
ok

pbiering:
Code validation is not working on updated PRs, currently also no clue how to trigger manually...
but extended documentation further on how to run locally: https://github.com/Kozea/Radicale/wiki/Contributions

````
radicale/app/report.py:72: error: Need type annotation for "props"  [var-annotated]
radicale/app/report.py:224: error: Item "None" of "Optional[Any]" has no attribute "serialize"  [union-attr]
radicale/tests/test_base.py:1550: error: Item "int" of "Union[int, dict[str, tuple[int, Element]]]" has no attribute "values"  [union-attr]
radicale/tests/test_base.py:1558: error: Need type annotation for "uids" (hint: "uids: List[<type>] = ...")  [var-annotated]
radicale/tests/test_base.py:1590: error: Item "int" of "Union[int, dict[str, tuple[int, Element]]]" has no attribute "values"  [union-attr]
Found 5 errors in 2 files (checked 73 source files)
```

Had to fix also  a report by `isort`, so please rebase before submitting further fix

metallerok:
@pbiering mypy gives strange results, here is the code from an existing test
```python
        self.put(event_path, event)
        _, responses = self.report(event_path, """\
<?xml version="1.0" encoding="utf-8" ?>
<C:calendar-query xmlns:C="urn:ietf:params:xml:ns:caldav">
    <D:prop xmlns:D="DAV:">
        <D:getetag />
    </D:prop>
</C:calendar-query>""")
        assert len(responses) == 1
        response = responses[event_path]
        assert not isinstance(response, int)
        status, prop = response["D:getetag"]
        assert status == 200 and prop.text
```

and here is the code from my test
```python
        self.put("/calendar.ics/", get_file_content("event_daily_rrule.ics"))
        req_body_without_expand = ...
        _, responses = self.report("/calendar.ics/", req_body_without_expand)
        assert len(responses) == 1

        response_without_expand = responses['/calendar.ics/event_daily_rrule.ics']
        status, element = response_without_expand["C:calendar-data"]

        assert status == 200
```

But mypy doesn’t respond to the first section of the code, and to mine it gives an error, not understanding that the result is being unpacked.

And another question: should I add validation fixes to the current branch or to a new one?

```
> .venv/bin/python -c 'import importlib.util, subprocess, sys; importlib.util.find_spec("mypy") and sys.exit(subprocess.run(["mypy", "."]).returncode) or print("Skipped: mypy is not installed")'
```

```
radicale/tests/test_base.py:1551: error: Value of type "Union[int, Dict[str, Tuple[int, Element]]]" is not indexable  [index]
radicale/tests/test_base.py:1555: error: Unsupported right operand type for in ("Union[Any, str, None]")  [operator]
radicale/tests/test_base.py:1556: error: Unsupported right operand type for in ("Union[Any, str, None]")  [operator]
radicale/tests/test_base.py:1557: error: Unsupported right operand type for in ("Union[Any, str, None]")  [operator]

```

image

PR #1305, opened 2023-06-06
changed rel="shortcut icon" to rel="icon"

heull001:
"icon" is standard and used by all modern browsers. Only IE needed "shortcut icon". Since is it not longer supported and would have needed an .ico-file, "icon" is the right value for rel.


pbiering (approved):
ok

Issue #1210, opened 2022-01-26
Radicale in root folder no longer working 


metlinux:
After upgrading to Radicale 3.1.2, Radicale stopped working for me with the following error, returning code 400
  [ERROR] Base prefix (from HTTP_X_SCRIPT_NAME) must not end with '/': '/'

Reading the release notes, I am pretty sure this is because of the this point that is mentioned in the release notes:
  - Verify that base prefix starts with '/' but doesn't end with '/'

In my case I am running radicale on a seperate domain, without a path prefix with the following apache reverse proxy settings
```
   <Location "/">
        AllowOverride None
        ProxyPass        http://localhost:5232/ retry=0
        ProxyPassReverse http://localhost:5232/
        RequestHeader    set X-Script-Name /
        RequestHeader    set X-Remote-User expr=%{REMOTE_USER}
    </Location>
```

So with just using an url like https://dav.example.com and no subfolders. This means my prefix starts en ends with / because that is all I use and I need to enter something as X-Script-Name.
This has been working like for ever ;-) but now with the new requirement for the prefix I run into this error caused by the "RequestHeader set X-Script-Name /"

In what way would I be able to continue using Radicale in the root folder of my domain ? So without a path prefix.
It will be cumbersome to alter all clients with a new (and extended) URL and I prefer to continue working without a prefix

Unrud:
Just remove the line `RequestHeader    set X-Script-Name /` or alternatively set it to an empty string.

metlinux:
> Just remove the line `RequestHeader set X-Script-Name /` or set it to an empty string `RequestHeader set X-Script-Name ""`.

That indeed is all that is required. Thanks Unrud!

pbiering:
what was the reason for adding that line at all?

```
RequestHeader    set X-Script-Name /
```

PR #1310, opened 2023-06-15
Document removal of X-Script-Name when serving Radicale at root (/)

quite:
See https://github.com/Kozea/Radicale/issues/1210

If X-Script-Name header is still set to `/` we get warnings in the log


pbiering (approved):
ok

PR #1320, opened 2023-08-05
Add example configuration for Caddy web server

jummo:
Add an example configuration for Caddy web server to DOCUMENATION.md file.


pbiering (approved):
ok

PR #1328, opened 2023-08-30
Remove brackets from IPv4 test hostnames

tjni:
Python 3.11 follows the URI spec more closely and no longer allows braces surrounding IPv4 addresses in hostnames. This is a side effect of https://github.com/python/cpython/issues/103848.


pbiering (approved):
ok

Issue #1316, opened 2023-07-07
Worker thread crashes if a VCALENDAR component contains both VEVENT and VJOURNAL components


rhymeswithmogul:
I noticed an odd bug in Radicale 3.0.6.  When it encounters an iCalendar file that contains both a VEVENT and VJOURNAL item inside the VCALENDAR object, it fails to process it.  

```
[129269/Thread-2395 (process_request_thread)] [ERROR] An exception occurred during REPORT request on '/colin/MyCalendarUUID/': Failed to load item 'MyEventUUID.ics' in 'colin/MyCalendarUUID': Multiple component types in object: 'VEVENT', 'VJOURNAL'
```

However, this does appear to be valid according to RFC 5545, Section 3.6, which says, "The body of the iCalendar object consists of a sequence of calendar properties and one or more calendar components".

I admit that I did add this VJOURNAL section to an .ics file by hand, to test some code that I'm writing.  The file looked something like this:

```icalendar
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//mozilla.org/Mozilla Thunderbird//EN
BEGIN:VTIMEZONE
[timezone stuff here]
END:VTIMEZONE
BEGIN:VEVENT
[event stuff here]
END:VEVENT
BEGIN:VJOURNAL
[journal items here]
END:VJOURNAL
END:VCALENDAR
```

I copied this file into two files, one with the VJOURNAL components removed and one with the VEVENT component removed, and Radicale synced it correctly.

pbiering:
this is not an issue of `radicale` itself but the underlying object parser library `vobject` (https://github.com/eventable/vobject). Unfortunatly this is turned stale. One has to investigate which of the forks is actively fixing bugs on-top and would be a best candidate to file such issue (and potentially replace "upstream" if it stays stale forever).

rhymeswithmogul:
Got it. Thank you.

da4089:
@rhymeswithmogul, I attempted to replicate this issue today.  I used to following ICS file (cobbled together from other vObject test cases):

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//mozilla.org/Mozilla Thunderbird//EN
BEGIN:VTIMEZONE
TZID:US/Pacific
BEGIN:STANDARD
DTSTART:19671029T020000
RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=10
TZOFFSETFROM:-0700
TZOFFSETTO:-0800
[...]

PR #1370, opened 2024-03-02
fix for Worker thread crashes if a VCALENDAR component contains both VEVENT and VJOURNAL components

pbiering:
https://github.com/Kozea/Radicale/issues/1316

Replace code which is not permitting one-time occurance of VEVENT and VJOURNAL by using a dictionary.

Also had to remove the duplicate UID verification, because VEVENT and VJOURNAL have different UID usually.


pbiering:
this breaks somehow https://github.com/Kozea/Radicale/blob/master/radicale/tests/static/event2.ics, have to investigate next

pbiering:
broken

PR #1421, opened 2024-03-11
WEB UI: Hide download button for Webcal collections [SMALL]

MatthewHana:
Hides the download button for Webcal collections in the Web UI (which have no content to download)


pbiering (approved):
ok

PR #1424, opened 2024-03-11
Add custom Radicale DAV property getcontentcount

MatthewHana:
This adds a custom Radicale DAV property called 'getcontentcount' which returns the number of entries contained within that collection.

### DIFF
--- a/radicale/app/propfind.py
+++ b/radicale/app/propfind.py
@@ -272,6 +272,12 @@ def xml_propfind_response(
                     element.text = displayname
                 else:
                     is404 = True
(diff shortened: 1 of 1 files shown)


pbiering:
@MatthewHana : please fix the failing checks

MatthewHana:
I'm not sure why the tests are failing. Something to do with the mypy dependency.

MatthewHana:
Tests all passing now.

Original code was fine. Static type checking test kept causing an issue about a scenario that could not happen due to an earlier check. Added the check for a second time in the block to satisfy it.

PR #1429, opened 2024-03-12
fix for pytest warning: "is using nose-specific method: `setup(self)`"

pbiering:
- `teardown` -> `teardown_method`
- `setup` -> `setup_method`


pbiering:
looks like this is incompatible with older Python versions (e.g. 3.9) causing now troubles like

```
TypeError: too many positional arguments
```

How can this be conditionally adjusted?

PR #1473, opened 2024-04-22
Ability to create predefined calendar or(and) addressbook for new user

itglob:


DOCUMENTATION.md
pbiering, line 1: for what is the added "Default:"?

config
pbiering, line 1: from this I can't see how to define the option proper

radicale/app/__init__.py
pbiering, line 1: the feature should be disabled by default if the property is not set, how is this covered here?

pbiering:
Good idea, but please check my comments.

radicale/app/__init__.py
itglob: disabled by default
json_str return {} (empty dict) 
radicale/app/__init__.py  - line 275  - empty loop

config
itglob: is my example in documentation and config unclear?

DOCUMENTATION.md
itglob: "Default:" copied from "hook".
remove ?

itglob:
disabled by default
json_str return {} (empty dict)
```radicale/app/__init__.py``` - line 275 - empty loop


is my example in documentation and config unclear?

pbiering:
I had no time to test so far, but scheduled

pbiering:
Can you please extend the code that it validates the config option during start? It would be helpful to avoid any broken config sniplet detected only on the "new user" trigger.

itglob:
Which aspect should I check? The json syntax is checked during json.loads in config.py.

pbiering:
> Which aspect should I check? The json syntax is checked during json.loads in config.py.

Best that they are valid collection properties (calendar or address book).

Alternative: hardcode and only add an option for enabling of that feature, then the QA can be done by test script.

itglob:
check_and_sanitize_props called before create_collection in ```app/__init__.py - l277```
i didn't think we need do this in config.

pbiering:
> check_and_sanitize_props called before create_collection in `app/__init__.py - l277` i didn't think we need do this in config.

but this will be triggered on first usage only, and then it breaks....I would expect that "preloaded" config is assured working and not waiting to fail on first usage.

PR #1476, opened 2024-04-27
Allow quoted-printable encoding for vObjects.

da4089:
Resolves the parsing issue in #1238.

Issue #1238, opened 2022-04-25
radicale upload VCARD2.1 failed


osnosn:
In debian11, `apt install radicate` installed radicate-3.0.6.

when I upload a VCF file via browser, it failed.

This is the VCF file:   
```
BEGIN:VCARD
VERSION:2.1
N;CHARSET=UTF-8;ENCODING=QUOTED-PRINTABLE:;=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=
=95=E6=B5=8B=E8=AF=95=E6=B5=8B;;;
FN;CHARSET=UTF-8;ENCODING=QUOTED-PRINTABLE:=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=
=95=E6=B5=8B=E8=AF=95=E6=B5=8B
TEL;CELL:12 345 678 90
PHOTO;ENCODING=BASE64;JPEG:/9j/4AAQS....
 2wBDAQcHBwo....

END:VCARD
``` 

------

I changed VCF file, then upload OK.   
* change QUOTED-PRINTABLE in one line.
* change PHOTO:ENCODING=BASE64; to PHOTO:ENCODING=B; 
```
BEGIN:VCARD
VERSION:2.1
N;CHARSET=UTF-8;ENCODING=QUOTED-PRINTABLE:;=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B;;;
FN;CHARSET=UTF-8;ENCODING=QUOTED-PRINTABLE:=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B=E8=AF=95=E6=B5=8B
TEL;CELL:12 345 678 90
PHOTO;ENCODING=B;JPEG:/9j/4AAQS....
 2wBDAQcHBwo....

END:VCARD
```

Barugon:
I've been trying to upload my contacts (vCard 3.0) and I get the same thing.

slyduda:
I was able to fix this by commenting the following lines from the vobject package `vcard.py` file 

```py
[...]

PR #1477, opened 2024-04-28
Add a test case for #1238

da4089:
The key is the QP continuation line for the URL, which (unlike vCard) doesn't indent, but uses a trailing '='.  Parsing this needs the vobject.readComponents() call to have its allowQP flag True, as implemented in #1476.


pbiering (approved):
ok

pbiering:
Thank you!

PR #1488, opened 2024-05-06
Fix missing icons if installed via pip

Rilele:
The icons for the new WebUI are not included in web_files in setup.py . Because of this, the icons are missing if Radicale is installed via pip.

This pull request fixes that.


nebulade:
Came here to report the same while updating the Cloudron package.  I can confirm this patchset fixes the issue. Nearly submitted the same https://github.com/cloudron-io/Radicale/commit/63a422f13f1de259e2202e3274045ce3bdacee4c

pbiering (approved):
thank you

PR #1489, opened 2024-05-08
Update pytest

### LINKED ISSUE #1486: Pytest 8 support
Hi! :wave: 

I'm packaging this project for Arch Linux.
We are currently trying to get rid of pytest 7 in our repositories, which we had to introduce for some packages.
Radicale appears to be not compatible with pytest > 7 (which is now at 8.1.2) and also locks typeguard to < 3 (which is now at 4.2.1).

It would be awesome if compatibility could be fixed! :)
Thanks

Unrud:
Closes #1429
Fixes #1486

### DIFF
--- a/radicale/tests/__init__.py
+++ b/radicale/tests/__init__.py
@@ -48,7 +48,7 @@ class BaseTest:
     configuration: config.Configuration
     application: app.Application
 
-    def setup(self) -> None:
+    def setup_method(self) -> None:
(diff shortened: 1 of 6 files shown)


pbiering (approved):
thank you very much!

pbiering:
Switching from "typeguard<3" to "typeguard>=4" causes problems now for all tests excluding "*-pypy*" :-(

Issue #1485, opened 2024-05-03
datetime.date object has no attribute astimezone


10b14224cc:
I get the following error with the new version of radicale:
```
[2024-05-04 01:57:32 +0200] [539526/Thread-80 (process_request_thread)] [ERROR] An exception occurred during REPORT request on '/raffaele/901469ab-d739-9859-86bd-6f0c1334d5b2/': 'datetime.date' object has no attribute 'astimezone'
```

SpiritCroc:
I needed a quick fix for this, so I went into `app/report.py` and changed the place where `_expand()` was called to a try-except, i.e.

```
try:
    expanded_element = _expand(
        element, copy.copy(item), start, end)
    found_props.append(expanded_element)
except Exception as e:
    logger.warning("Expand failed: %s ", e)
    found_props.append(element)
```

If I understand the issue correctly, 3.2.0 added support for expanding recurring events, but hasn't considered full-day events stored as `date` instead of `datetime`, so a proper fix should change the logic to account for that.

pbiering:
@metallerok : can you please dig into, potentially https://github.com/Kozea/Radicale/pull/1293 was incomplete

pbiering:
> @metallerok : can you please dig into, potentially #1293 was incomplete

@metallerok : did you have already time to look into?

metallerok:
yes, i am trying

pbiering:
@SpiritCroc / @10b14224cc : can you please test latest master now?

SpiritCroc:
Thanks for looking into this!
With master, now the query doesn't return a server error anymore, which is great! However, the events appear to no longer be reported as full-day events. At least [my calendar script](https://gitlab.com/SpiritCroc/matrix_daily_calendar) used to say something like

"all day 2024-05-28", now it says something like
"2024-06-11 02:00 - 02:00"
which I assume comes from living in UTC+2, but would be nice if the event was still reported as "full day" (not sure if caldav has a flag for whole-day events or one needs to omit times or how this is decided under the hood).

metallerok:
Do you mean that expanded events return datetime, not date? Expanded events return the UTC dateime in accordance with the rfc [https://www.rfc-editor.org/rfc/rfc4791#section-9.6.5](https://www.rfc-editor.org/rfc/rfc4791#section-9.6.5)

metallerok:
[...]

PR #1505, opened 2024-06-02
(#1485) If an event comes with a dtstart specified as a date then in …

metallerok:
if an event comes to us with a dtstart specified as a date then in the response we return the date, not datetime


pbiering (approved):
ok

PR #1517, opened 2024-06-09
Python 3.13

pbiering:


pbiering:
"ubuntu-latest" has still no Python 3.13 preview on board...will be postponed...

deronnax:
I think the correct naming would be '3.13.0-beta.4'.
Source: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#python, especially the 2nd bullet point: *Downloadable Python versions from GitHub Release* , which led me to https://github.com/actions/python-versions/releases, which led me to https://github.com/actions/python-versions/releases/tag/3.13.0-beta.4-9993951762.

pbiering:
@deronnax: thank you for the hint...so far the test runs failing :-(

pbiering:
> Yep, still the same python 3.12.4+ error: `TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'`. We need to move to a more recent version of typeguard because it's [fixed in 4.3.0](https://github.com/agronholm/typeguard/releases/tag/4.3.0) (in [agronholm/typeguard#460](https://github.com/agronholm/typeguard/pull/460)).
> 
> Btw, is typeguard really used?

I don't know, haven't investigated so far. But I had to pin to 4.2.1 as 4.3.0 throwed other errors afair.

> I saw almost no occurrences in the code.

One need to check, potentially historic purposes?!

Issue #1282, opened 2023-02-26
Poor performance after server running for a few days


Terrance:
I'm seeing certain requests taking durations into the minutes to complete.  For example, loading my calendars in InfCloud:

|Method|Path|Duration|Note|
|-|-|-|-|
|PROPFIND|/|16 ms|
|REPORT|/user/|11 ms|
|PROPFIND|/user/|13 ms|Reading server settings|
|PROPFIND|/user/|**4.64 min**|Reading calendar / address book list|
|REPORT|/user/1a518a80-23b6-abd1-cef8-f0e25e6063ef/|**2.43 min**|Reading events|
|PUT|/user/1a518a80-23b6-abd1-cef8-f0e25e6063ef/cca...892.ics|**1.22 min**|Creating an event|

I've bumped my nginx proxy timeout to keep the requests working, though it makes things difficult to use with the long delays.

Performance seems to vary somewhat (e.g. give or take a few minutes for a scan of all collections) from day to day despite no obvious difference in server load.  Restarting Radicale helps (e.g. reducing the scan time to about 30 seconds initially) for a day or two, but the slowness creeps back in after that.

I'm not sure if I'm at the point where my calendars are just too hefty to work with as files, or if there's another bottleneck here.

Debug log snippet: [radicale-2023-02-26.log](https://github.com/Kozea/Radicale/files/10833737/radicale-2023-02-26.log) (NB. different day to the table)

<details>
<summary>Collection stats</summary>

For context, as I'm not sure what constitutes a "large" collection, here's some stats on the collection root (I'm the only user):

|Collection|Type|# items|Total size|
|-|-|-|-|
|1a518a80-23b6-abd1-cef8-f0e25e6063ef|VCALENDAR|12836|651M|
|1fc7a94b-a1b1-bb0a-5e9e-c3756f597434|VCALENDAR|921|14M|
|260ceee4-135a-69a8-d36c-8bfe3937ed2a|VADDRESSBOOK|155|6.0M|
|29917d2c-60e7-0a32-178b-079c5bb33fe8|VCALENDAR|3060|46M|
|31f4fd59-7748-b1de-426c-ee1bee515ef4|VCALENDAR|3052|47M|
|4c899125-a332-8e3d-ce5b-6aabb3db886a|VCALENDAR|25|348K|
|513bbed6-56f0-cd92-4188-5c3c076745a5|VCALENDAR|167|2.3M|
[...]

PR #1531, opened 2024-07-01
Replace has_uid with Filename Check for Performance Improvements

koalyorg:
**Issue:**
When using Radicale with a larger number of items, the system becomes very slow (#1282). This performance issue arises because, on each PUT/MOVE operation, the has_uid function is called. This function loads all items (self.get_all()) to check if the specified UID is unique.

**Solution:**
Instead of loading all items, this patch ensures that the filename matches the UID, thereby ensuring that the UIDs remain unique without the need to load all items for such.

**Changes:**
- Removes has_uid function
- Add check if filename exists with <UID>.vcf in PUT and MOVE

**Improvements**
- Significant performance improvement for large collections when performing PUT or MOVE.

Please review the changes and provide feedback. Thank you.

### DIFF
--- a/radicale/app/move.py
+++ b/radicale/app/move.py
@@ -3,6 +3,7 @@
 # Copyright © 2008 Pascal Halter
 # Copyright © 2008-2017 Guillaume Ayoub
 # Copyright © 2017-2018 Unrud <unrud@outlook.com>
+# Copyright © 2024 koalyorg <info@koaly.org>
 #
 # This library is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


pbiering:
@koalyorg : this is a major behavior change because it makes mandatory that file name without extension matches item.uid inside. At least all test items in `radicale/tests/static/` have to be reviewed.

Is there any RFC mention this as "has-to-be" (anyhow)?

Please add an option (default=off) to conditional enable that performance improvement.

BTW: before approval, please run the whole test suite by yourself, see https://github.com/Kozea/Radicale/wiki/Development-Testing

Another option would be to create a memory cache on-top of the file system cache covering only the item.uid per collection.

koalyorg:
@pbiering You are right; this is a significant behavior change. After reviewing the RFC in more detail, it seems the proposed solution is not ideal, as there is no fixed way for naming the files, at most a possible convention to include the UID in the name. Therefore, caching is likely the only viable option to address this issue in a file-based setup. Unfortunately, I won't have the time to look into this.

PR #1541, opened 2024-07-12
Modified Reverse-Proxy Examples for Caddy

mldytech:
- Moved existing Caddy example with basicauth to the right section
- Added basic Caddy example without using basicauth
- Changed radicale* radicale/*, following the convention


pbiering (approved):
ok

PR #1545, opened 2024-07-23
remove setuptools from runtime dependencies

deronnax:
You don't need setuptools as a runtime dependency.


pbiering (approved):
Thank you!

PR #1546, opened 2024-07-23
Migrate to pyproject.toml

deronnax:
It's the future. The setup.cfg needs to stay because[ flake8 refuses to migrate to pyproject.toml](https://github.com/PyCQA/flake8/issues/234). The `tox` config stays there as well, as tox config in [pyproject.toml is currently basically inlined tox setup.cfg config](https://tox.wiki/en/latest/config.html#pyproject-toml).


pbiering:
This change can cause issues in external packaging (e.g. RPM or DEB) which need to be potentially adjusted, postponed to 3.3.0 to avoid such potential unexpected change in a very minor version.

deronnax:
when is 3.3.0 coming ?

pbiering:
> when is 3.3.0 coming ?

only code-related outstandings are this and your other PR, but these can be postponed to 3.3.x if you wish

deronnax:
let's get it merged for 3.3.0

pbiering:
> regarding `requires = ["setuptools>=61.2"]` 

Hmm, saw now that meanwhile Python 3.11 arrived in EL8 and EL9

```
python3.11-setuptools.noarch    65.5.1-3.el8_10    appstream  
python3.12-setuptools.noarch    68.2.2-4.el8_10    appstream  

python3.11-setuptools.noarch    65.5.1-2.el9_4.1   appstream
python3.12-setuptools.noarch    68.2.2-3.el9_4.1   appstream
```

So in case this PR is applied, packaging for EL8+9 has to change then from Python 3.9 to 3.11 - should be not a big deal.

pbiering:
> what is EL8 ?

Enterprise Linux major version 8 - Red Hat, Alma, Rocky...

pbiering:
@deronnax : please rebase to latest, currently your branch has conflicts

pbiering:
@deronnax can you please dig into why the "lint" test failed?

pbiering:
@deronnax this broke more now...can you try to run test suite locally (https://github.com/Kozea/Radicale/wiki/Development-Testing-Release#complete-test-suite) to fix this

pbiering (approved):
ok

pbiering:
@deronnax - is it possible to provide still setup.py for legacy support? I ran here into a **deadlock** while trying packaging for EPEL9:

- pyproject.toml requires a higher version than provided with EL9 python39 (python3-setuptools-53.0.0-12.el9_4.1.noarch)
- switching to Python 3.11 (and using python3.11-setuptools-65.5.1-2.el9_4.1.noarch) would work in general, but looses support of
  - "bcrypt" (as there is no RPM available, missing dependency python3.11-bcrypt)
  - "pika" (as there is no RPM available, missing dependency python3.11-pika)

and "bcrypt" cannot even bundled, because it contains a binary, while "radicale" is noarch RPM which cannot include binaries.

pbiering:
BTW: in worst case I revert this until at least python3.11-bcrypt appears on EPEL9.

pbiering:
Just note that I was able to use the legacy files to package for EL8 and EL9, will include them in the package for now with .legacy suffix and in RPM spec file I will reactivate them for EL8 and EL9 build.

deronnax:
we could have migrate the setup.py to setup.cfg, which was the "previous" standard before pyproject.toml. If you want I can do that and this way we still have declarative config but it will work on you old EL.

pbiering:
@deronnax - currently I have adjusted in pre-next release tests simply the spec files for EL8 and EL9 and use the setup.*.legacy files - working well:

```
%{__rm} pyproject.toml
%{__cp} setup.cfg.legacy setup.cfg
%{__cp} setup.py.legacy setup.py

...
%py3_build
...
%py3_install
```

Fedora spec  files already adjusted for using pyproject.toml - which was a minor change here only.

If this can be used in this manner for the next time, no further adjustments are needed at the moment and nothing is blocking the 3.3.0 release planned for today.

Also other packagers for "legacy" / "old" distributions can use the same mechanism if required.

PR #1547, opened 2024-07-24
fix misspellings

deronnax:
Some misspellings I found. Misspellings can prevent code grep and efficient logs exploitation.


pbiering (approved):
Thank you very much!

PR #1538, opened 2024-07-12
Workflow test

pbiering:

PR #1548, opened 2024-07-24
split tox jobs

deronnax:
You don't need to run isort, mypy and flake8 as many time as you have environments (~20 here).
* you save yourself Github Actions CI minutes (the Windows one count double and macOS count 4 times)
* you get faster feedback (the ubuntu 3.11 test one runs under 20 seconds)(we could probably stop the test matrix as soon as this one fail, to avoid useless runs).
* no more dirty python / mypy version check hack.
* you save the planet

Btw, mypy, isort and flake8 version should be pinned. If not, you are at risk than any new release of any of those 3 turns your CI red overnight.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install Test dependencies
         run: pip install tox
       - name: Test
-        run: tox
+        run: tox -e py
       - name: Install Coveralls
         if: github.event_name == 'push'
         run: pip install coveralls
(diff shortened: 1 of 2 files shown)


pbiering:
> Btw, mypy, isort and flake8 version should be pinned. If not, you are at risk than any new release of any of those 3 turns your CI red overnight.

can you add this also to this PR (similar happened already in the past with "Python 3.12.3->3.12.4")

https://github.com/Kozea/Radicale/pull/1538/files

and related to "tox":

```
[tox:tox]
min_version = 4.0
```

Is there also a `max_version` supported?

deronnax:
Pinning done. And no, there is no tox `max_version`. Why so?

pbiering:
> Pinning done. 

Thank you!

> And no, there is no tox `max_version`. Why so?

Avoid that e.g. 5.0 causes broken builds...but potentially "tox" is not "playing the game" with developers as others do by introducing breaking changes or add new strong dependencies in even sub-minor releases.

pbiering (approved):
thank you

PR #1553, opened 2024-08-06
pyupgrade --py38-plus

deronnax:
Mostly remove dead python < 3.8 code.

Btw, what is the python support policy ? When is python 3.8 support dropping ?

### DIFF
--- a/radicale/config.py
+++ b/radicale/config.py
@@ -349,7 +349,7 @@ def load(paths: Optional[Iterable[Tuple[str, bool]]] = None
         config_source = "config file %r" % path
         config: types.CONFIG
         try:
-            with open(path, "r") as f:
+            with open(path) as f:
                 parser.read_file(f)
                 config = {s: {o: parser[s][o] for o in parser.options(s)}
                           for s in parser.sections()}
--- a/radicale/server.py
+++ b/radicale/server.py
@@ -176,7 +176,7 @@ class ParallelHTTPSServer(ParallelHTTPServer):
             if name == "certificate_authority" and not filename:
                 continue
             try:
-                open(filename, "r").close()
+                open(filename).close()
             except OSError as e:
                 raise RuntimeError(
                     "Invalid %s value for option %r in section %r in %s: %r "
--- a/radicale/types.py
+++ b/radicale/types.py
@@ -15,9 +15,9 @@
 # along with Radicale.  If not, see <http://www.gnu.org/licenses/>.
 
 import contextlib
-import sys
 from typing import (Any, Callable, ContextManager, Iterator, List, Mapping,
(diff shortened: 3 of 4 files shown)


deronnax:
wtf pypy segfault

pbiering:
> Btw, what is the python support policy ? When is python 3.8 support dropping ?

Python 3.8 has EOSL 2024-10 - I assume the for github workflows used template will disappear afterwards.

For Enterprise Linux 8 since 3.2.0 Python 3.9 is selected during packaging, so support for 3.8 is not mandatory anymore.

See also:
https://github.com/Kozea/Radicale/wiki#tested-python-versions

pbiering:
> Mostly remove dead python < 3.8 code.

I would not remove "dead" code so far unless complains are received that existing version is not running with < 3.8 at all. Current status is that it cannot be tested anymore using github workflow.

BTW: there is also a poll available: https://github.com/Kozea/Radicale/discussions/1408

pbiering:
some tests failing, can you please check?

deronnax:
I'll take care, yes. Regarding python 3.8: if it can't be tested anymore, it is now probably broken, so don't bother and remove the dead code.

pbiering:
> Regarding python 3.8: if it can't be tested anymore, it is now probably broken, so don't bother and remove the dead code.

For now Python 3.8 should be still testable hopefully

deronnax:
I meant: python < 3.8, sorry. Currently, due to [this line in the setup.py](https://github.com/Kozea/Radicale/blob/master/setup.py#L63), Radicale will refuse to install on python 3.7 and older. Python 3.7 has been released 6 years ago and is end of life for 1,5 years now. Even without removing dead code, after this MR is merged, it's pretty sure Radicale won't run anymore on 3.7. If you are not OK with this, just close this MR.

pbiering:
> I meant: python < 3.8, sorry. Currently, due to [this line in the setup.py](https://github.com/Kozea/Radicale/blob/master/setup.py#L63), Radicale will refuse to install on python 3.7 and older. Python 3.7 has been released 6 years ago and is end of life for 1,5 years now. Even without removing dead code, after this MR is merged, it's pretty sure Radicale won't run anymore on 3.7. If you are not OK with this, just close this MR.

Support of Python 3.7 ended with 3.1.9, see also https://github.com/Kozea/Radicale/wiki

Also test suite is not testing 3.7 anymore, but 3.u triggers still an issue

https://github.com/Kozea/Radicale/actions/runs/10265934628/job/28403176547?pr=1553

It's a segfault, somehow strange, potentially a glitch, can you commit a more or less empty change to be able to retrigger the tests?

deronnax:
we have the answer: it was a glitch. It can be merged.

Issue #1490, opened 2024-05-13
Disable  hook "rabbitmq" with warning on startup in case required "pika" is not available


pbiering:
this is related to https://bugzilla.redhat.com/show_bug.cgi?id=2280233 where I have overseen, that "pika" is a requirement for 3.2.0.

While not a problem on Fedora or EPEL9 this causes a problem with the limited amount of Phyton packages for versions 3.8 and 3.9.

=> code should be adjusted to to avoid dependencies to Python modules for only optional features.

pbiering:
@freakmaxi : as this is strongly related to https://github.com/Kozea/Radicale/pull/1092/files can you please work on an extension which in case of the required modules are not available on the system simply drop a warning (feature not supported) and disable all related code parts?

freakmaxi:
Hello @pbiering 

I'm currently on my Annual vacation leave and I return back at the end of June. I'll take care of it after my return. How urgent to complete the task?

pbiering:
@freakmaxi : no hurry here anymore, I was able to fix the issue in the EL8 RPM by adding an additional "__requires_exclude".

But at least the code should be reviewed in case an option is selected which requires "pika", that it should simply test this during startup and add a note about "not supported" and disable the uuse. I assume currently it will throw an execption.

henning-schild:
Note that pika does not officially support python 3.12, or at least does not provide a statement on its support status.

Without such a statement from upstream distros will likely not want to package it when going to 3.12, which would consequently leave radicale stuck at <3.2.0. That is currently the state in gentoo Linux, we can not bump to 3.2.1, we could jump to something that does not need pika.

https://bugs.gentoo.org/929480

henning-schild:
In fact not installing pika at all does not make the testsuite fail, which indicates questionable test coverage on the whole hook code.

pbiering:
@freakmaxi : please dig into.

@henning-schild : all dependencies currently used are on risk on any minor Python version update...already next known upcoming: "passlib" (looks like unmaintained since 2020), see https://github.com/Kozea/Radicale/issues/1405

PR #1554, opened 2024-08-06
hook: gracefully ignore non functional hooks and fall back to none

henning-schild:
In case a hook fails to load for some reason, fall back to the default hook "none" and treat errors as warnings in the log.

This will gracefully ignore typos in hook names without crashing the server, and it will also allow configuration of "rabbitmq" where i.e. "pika" is missing.

Closes: #1490


pbiering (approved):
thank you!

Issue #197, opened 2014-08-08
WSGI or similar deployment with Nginx


chlarsen:
Dear All,
I have successfully installed Radicale 0.8, and am very impressed swith its size, speed and functionality. I managed to get somewhat better performance by disabling DNS lookups, und use LDAP at Radicale application level for user authentication. The whole thing sits behind an Nginx reverse proxy. While I am patiently waiting for the PostgreSQL backend to work proprely (hopefully with the next stable release?), I got stuck with donig any further (accelerating) magic with WSGI (or FastCGI) and Nginx. An documentation? Alternative, I have used gunicorn with great success in the Django domain- could this work here instead: Radicale -> Gunicorn -> Nginx reverse proxy?
Thanks a lot for this terrific peice of software, and your kind assistance.
Chris

PR #1570, opened 2024-09-17
Fix issue #197

petervarkoly:
# Problem
[ERROR] An exception occurred during GET request on '/.web/': string indices must be integers, not 'str' when using LDAP

# Resolution
Fix using search_scope in ldap3 connection.

Additionaly enhancing debug logging.


pbiering (approved):
ok

PR #1576, opened 2024-09-22
Implementing group calendars and increase perfomance

petervarkoly:
### Implementing rights based only on group membership
I've adapted rights/from_file.py to be able to handle sections without user pattern.

### Increase perfomance
In rights/from_file.py the configuration file was opened and parsed by all call of  authorization. I've moved the opening and parsing of the configuration file into __init__ to increase the performance of the server. As side effect the server must be restarted when the rights configuration was changed.

### Implementing group calendars
I've enhenced the discover function in StoragePartDiscover class to handle group calendars. A new for loop was introduced which walks on the user_groups an check if a group calendar directory does exists. In case of yes an additional collection will be yield.

- The group calendar will be placed under collection_root_folder/GROUPS.
- The name of the calendar directory is the base64 encoded group name.
- The group calneder folders will not be created automaticaly. This must be created manualy. I'll provide a script on wiki.

### DIFF
--- a/config
+++ b/config
@@ -59,7 +59,7 @@
 # URI to the LDAP server
 #ldap_uri = ldap://localhost
 
-# The base DN of the LDAP server
+# The base DN where the user accounts have to be searched
 #ldap_base = ##BASE_DN##
 
 # The reader DN of the LDAP server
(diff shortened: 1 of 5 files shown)


pbiering (approved):
ok

pbiering:
@petervarkoly can you please also create PR for updating DOCUMENTATION.md

PR #1577, opened 2024-09-22
Draft: Allow reading from environment variables for options

poperigby:
I need to test this but it should work.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -105,6 +105,18 @@ You should create a new configuration file at the desired location.
 (If the use of a configuration file is inconvenient, all options can be
 passed via command line arguments.)
 
+Configuration options can also be passed through environment variables in the form of:
+
+```bash
+RADICALE_OPTION_<SECTION>_<OPTION>=<value>
+```
+
+For example:
+
+```bash
+RADICALE_OPTION_SERVER_HOSTS="0.0.0.0:5232, [::]:5232"
+```
+
 All configuration options are described in detail in the
 [Configuration](#configuration) section.
 
--- a/radicale/__main__.py
+++ b/radicale/__main__.py
@@ -149,9 +149,13 @@ def run() -> None:
     env_var_config: types.MUTABLE_CONFIG = defaultdict(dict)
     for key, value in os.environ.items():
         if key.startswith("RADICALE_OPTION_"):
-            # Split only on the first instance of _
-            section, option = key.removeprefix("RADICALE_OPTION_").split('_', maxsplit=1)
-            env_var_config[section.lower()][option.lower()] = value
+            try:
+                # Split only on the first instance of _
+                section, option = key.removeprefix("RADICALE_OPTION_").split('_', maxsplit=1)
+                env_var_config[section.lower()][option.lower()] = value
+            except ValueError as e:
+                logger.critical(f"Invalid environment variable format (Proper format is `RADICALE_OPTION_<SECTION>_<OPTION>): {e}")
+                sys.exit(1)
 
     # Update Radicale configuration according to command line arguments
     arguments_config: types.MUTABLE_CONFIG = defaultdict(dict)
--- a/radicale/__main__.py
+++ b/radicale/__main__.py
@@ -150,15 +150,14 @@ def run() -> None:
     for key, value in os.environ.items():
         if key.startswith("RADICALE_OPTION_"):
             # Split only on the first instance of _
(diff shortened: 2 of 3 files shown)


pbiering:
please provide update for DOCUMENTATION.md and also if possible create some test cases

poperigby:
@pbiering Will do. Where should I put the test, and also, would you like me to use `defaultdict(dict)` for `arguments_config` as well? Seems to be more elegant.

pbiering (approved):
ok

pbiering (approved):
ok

pbiering:
> @pbiering Will do. Where should I put the test

There is a test subdirectory, check whether you can copy+adjust one or more

poperigby:
Not quite sure how to do the test. I need to set an environment variable before the test server starts.

pbiering:
> Not quite sure how to do the test. I need to set an environment variable before the test server starts.

potentially there is a way to set environment variable in Python before calling the function.

poperigby:
The problem is that the setup function is called before any test is run I think. I was thinking of putting it in `test_server.py`.

pbiering:
> The problem is that the setup function is called before any test is run I think. I was thinking of putting it in `test_server.py`.

would it not possible to copy and adjust `test_command_line_interface` and instead of defining command line options, set environment by e.g. `os.environ[option] = value` before reaching the `subprocess.Popen`?

poperigby:
Yeah that should work. How can I then check Radicale's current configuration?

poperigby:
@pbiering I think it should be good now if you wanna take a look.

pbiering:
Generally it looks good but your test case leads me to that it would be not safe for options which have `_` inside.

My original proposal was like `RADICALE_OPTION_<section>__<key>="<value>"`

And looks like I had forgotten to mention the reason behind the `__` (2x `_`) separator was that `<section>` + `<key>` must be able to support single `_`. Dash `-` cannot be used because not supported e.g. in _bash_.

Please check and adjust/extend code and also test cases like

```
# option without separator
RADICALE_OPTION_server__hosts="<value>"

# option with separator
RADICALE_OPTION_auth__htpasswd_filename="<value>"
````

**Comment by poperigby** [comment]
Made a PR

**Review by pbiering — APPROVED** [review]
ok

**Comment by pbiering** [comment]
@poperigby : code approved but tests partially failed, please investigate, thank you!

**Comment by poperigby** [comment]
@pbiering That's bizarre I don't remember making the comment about making a PR. Must have been a mistake. As for using a double underscore, yeah I'll fix that. The test seems to be failing because the test environment is using an older version of Python that doesn't have `removeprefix()`?

**Comment by pbiering** [comment]
@poperigby : yes, `removeprefix()` was introduced with Python 3.9.0, but it's currently not planned to drop support for 3.8.0 already with Radicale 3.3.0 - can you try to circumvent this by catching the used Python version and create for < 3.9 a workaround code sniplet?

```
if sys.version_info >= (3, 9):
    ...current code...
else:
    ...workaround code...
```

**Comment by pbiering** [comment]
@poperigby while working on another issue it came into my mind why not using the command line options which fully support all all options of the config file and have precedence.

There are two ways support, example:

```
## simple variable
RADICALE_OPTIONS="--logging-level=debug --config=/etc/radicale/config --logging-request-header-on-debug --logging-rights-rule-doesnt-match-on-debug"; /usr/bin/radicale $RADICALE_OPTIONS

## as array
RADICALE_OPTIONS=("--logging-level=debug" "--config=/etc/radicale/config" "--logging-request-header-on-debug" "--logging-rights-rule-doesnt-match-on-debug"); /usr/bin/radicale $RADICALE_OPTIONS
```

Can this also help in your case, because then no extension would be required at all.

**Comment by poperigby** [comment]
Yeah that's definitely an option, although it's certainly not as nice looking or easy to write as environment variable versions of options.

**Comment by pbiering** [comment]
> Yeah that's definitely an option, although it's certainly not as nice looking or easy to write as environment variable versions of options.

there is also a 3rd method

```
RADICALE_OPTIONS=()
RADICALE_OPTIONS+=("--logging-level=debug")
RADICALE_OPTIONS+=("--config=/etc/radicale/config")
/usr/bin/radicale ${RADICALE_OPTIONS[@]}
```

requires no additional support and one can take the options directly from the online help and does not need to replace `-` by `_`

pbiering:
@poperigby are you satisfied already with the existing capabilities as described? Then your PR would not be needed.

poperigby:
Yes I suppose that would work.

PR #1579, opened 2024-09-23
Implementing ssl connection for ldap auth

petervarkoly:
3 new ldap parameters was introduced:
**ldap_use_ssl** to trigger the usage of ssl for ldap connection
**ldap_ssl_verify_mode** to set how the certificates need to be verified
**ldap_ssl_ca_file** he path to the CA file in pem format which is used to certificate the server certificate


pbiering (approved):
ok

PR #1595, opened 2024-10-15
fix casing in docker file

individual-it:
when building this docker threw a warning
`- FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 3)`
so this PR fixes that tiny thing


pbiering:
Thank you!

PR #946, opened 2019-04-30
Introduced dovecot authN module

galaxy4public:
This work is heavily based on prior art by Giel van Schijndel (#200),
however the code was made Python 3 compatible, error checking
has been improved, and tests were introduced:

```
$ python setup.py test --extras --addopts "-k auth"
[skipped]
Name                        Stmts   Miss Branch BrPart  Cover
-------------------------------------------------------------
radicale/auth/dovecot.py       76      2     26      2    96%
[skipped]
=== 34 passed, 16 skipped, 198 deselected in 2.35 seconds ===
```

PR #1610, opened 2024-10-30
Rebase galaxy4public patch on top of bf4f5834

sysnux:
Updated the dovecot pull request https://github.com/Kozea/Radicale/pull/946 from galaxy4public to be compatible with current Radicale3.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -746,6 +746,9 @@ Available backends:
 `ldap`
 : Use a LDAP or AD server to authenticate users.
 
+`dovecot`
+: Use a local Dovecot server to authenticate users.
+
 Default: `none`
 
 ##### htpasswd_filename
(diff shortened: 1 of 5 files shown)


pbiering:
@sysnux thank you very much for the rebased PR, but looks like test suite on Windows is not running:

```
ERROR    radicale:__init__.py:117 An exception occurred during PROPFIND request on '/': module 'socket' has no attribute 'AF_UNIX'
Traceback (most recent call last):
```

Best would be to exclude the auth/dovecot tests on Windows platform where it would be very unexpected that one runs here _dovecot_ in parallel.

sysnux:
You are right, unfortunately I don't know how to that.

pbiering:
> You are right, unfortunately I don't know how to that.

Can you try short after `def _test_dovecot(..` detect OS and skip (but log) the test.

Found a code sniplet which should be able to support test exclusion based on OS:

```
>>> import os

>>> os.name
'nt'

>>> import platform

>>> platform.system()
'Windows'
```

See also https://stackoverflow.com/questions/1854/how-to-identify-which-os-python-is-running-on

As I have no Python on Windows running, potentially some try and errors are required until working well.

sysnux:
Documentation added.
New try at skipping tests on Windows...

PR #1619, opened 2024-11-07
Expand overridden

### LINKED ISSUE #1618: Expand does not support overridden recurring events
When doing a report with expand, the server logic expands all recurring events, but it fails to take into account overridden recurrent events.

See the [RFC](https://datatracker.ietf.org/doc/html/rfc4791#section-7.8.3) for an example where "Event #2 bis" has changed (both in the summary as in the start time).

I will submit a PR based on #1617.

pieterhijma:
This PR is based on #1617, so this PR should be merged first. Currently it contains the commits of that PR as well.

Closes #1618.


pbiering:
@pieterhijma please investigate failed tests

pieterhijma:
Thanks for merging the other one. 

The test failed because I forgot to use module `typing` and because of that it failed on Python 3.8. 

The first commit adds the functionality and a test that is virtually the same as [RFC 4791 Sec. 7.8.3](https://datatracker.ietf.org/doc/html/rfc4791#section-7.8.3). 

I then took the liberty to factor out the expand tests from base in the second commit and in the last commit I refactored the expand test to reduce code duplication. I hope that's ok. In my opinion the comment from `test_base.py` "tests with simple requests" does not apply to report with expand. I can rearrange if necessary.

pbiering (commented):
ok

PR #1622, opened 2024-11-08
Expand taking timezone into account

### LINKED ISSUE #1621: Expand does not take timezones into account
Expand is supposed to report times in the UTC timezone and expand does exactly that. Currently, Radicale converts a recurring event defined in some timezone to UTC and then expands the recurring event. However, this is incorrect in the case where a recurring event is defined in some timezone that has daylight saving time and the recurring event crosses the daylight saving date. In that case, the user will see a time offset by an hour. 

For example, a user creates a weekly recurring event at 15:00 US/Eastern in standard time (no daylight saving). The user expects the meeting to be at 15:00 as well if daylight saving time is active. 

However, currently, Radicale would convert 15:00 US/Eastern standard time to UTC which is 20:00 and then expand this. All expanded start times would be 20:00 UTC which is incorrect in the period in which daylight saving time is active, because that would 16:00 US/Eastern. 

The solution is to expand the recurring event while taking timezones into account. After the recurring instances have been generated based on the timezone, the times should be converted to UTC.

I will submit a PR for this.

pieterhijma:
Closes #1621.


pieterhijma:
Note that the test case is essentially the example that was given in the issue.

pbiering (approved):
ok

Issue #1063, opened 2020-06-03
Invalid option 'protocol' in section 'server' in /etc/radicale/config


vishnuraghavb:
OS: Arch linux
Radicale --version
3.0.2

`radicale.service` fails to start with exit code=1

`journalctl` returned the following message:  
```
[CRITICAL] Invalid configuration: Invalid option 'protocol' in section 'server' in config file '/etc/radicale/config'
```
There was no issues with radicale until last update. The protocol value wasn't changed from the default though the service kept failing. 
```
[server]
...
# SSL Protocol used. See python's ssl module for available values
protocol = PROTOCOL_TLSv1_2
...
```
After commenting out the protocol. The service starts fine.

pbiering:
3.x is not supporting option `protocol` so far, but would be a good idea to implement, will investigate

PR #1624, opened 2024-11-14
SSL socket protocol + ciphersuite option

pbiering:
Add option to allow configuration of SSL protocol and ciphersuite


pbiering:
Supports https://github.com/Kozea/Radicale/issues/1063

PR #1641, opened 2024-12-02
Update DOCUMENTATION.md

claman:
remove unnecessary call for `proxy_ssl_trusted_certificate` in example SSL-enabled reverse proxy config

Per the [CertBot documentation](https://eff-certbot.readthedocs.io/en/latest/using.html#where-are-my-certificates), including `proxy_ssl_trusted_certificate` is only necessary if you're doing OCSP stapling, which requires some additional config that's probably outside the scope of radicale's documentation


pbiering (approved):
ok

PR #1665, opened 2024-12-25
Disable overloading BaseAuth login method

itglob:
Child classes of BaseAuth must overload _login method not login, otherwise overloading break functionality of {lc,uc}_username


pbiering (approved):
ok

PR #1669, opened 2025-01-03
More LDAP auth patches

marschap:
Hi,

here's another round of LDAP auth patches
They deal with issues not dealt with in the previous round:
* avoid code duplication by calculating attributes to query in `__init__()`
* allow using LDAP authentication also to servers that do not know the `memberOf` attribute
  by querying it only if `ldap_load_groups` is `True`
* introduce config option `ldap_groups_attribute` that allows us to configure the groups attribute name
  instead of having it being hard coded to `memberOf`
* get rid of `ldap_load_groups` config option. - having `ldap_groups_attribute` being unset
  resp. being set to a specific attribute name has the exact same effect
* stop fiddling around with getting the RDN value of the groups attributes returned, 
  and instead use the LDAP modules' helper functions foreseen for this purpose.
  In addition, this gives us a bit more flexibility with the `ldap_groups_attribute`: we can use attribute with non-DN syntax too.
  As radicale does not rely on on the DN syntax, but only takes the RDN value anyway, this is no limitation, but a real extension
  
  Please consider incorporating them into mainline radicale

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -941,6 +941,12 @@ Load the ldap groups of the authenticated user. These groups can be used later o
 
 Default: False
 
+##### ldap_groups_attribute
+
+The LDAP attribute to read the group memberships from in the user's LDAP entry if `ldap_load_groups` is True.
+
+Default: `memberOf`
+
 ##### ldap_use_ssl
 
 Use ssl on the ldap connection
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -926,26 +926,26 @@ The search filter to find the user DN to authenticate by the username. User '{0}
 
 Default: `(cn={0})`
 
-#### ldap_user_attribute
+##### ldap_user_attribute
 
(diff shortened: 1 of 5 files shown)


DOCUMENTATION.md
```diff
@@ -932,14 +932,20 @@ The LDAP attribute whose value shall be used as the user name after successful a
 
 Default: not set, i.e. the login name given is used directly.
 
-##### ldap_load_groups
+#### ldap_groups_attribute
```
pbiering, line 935: should have 5x #, please fix also current buggy entry:

```
#### ldap_user_attribute
```

pbiering: Please also extend the changelog

marschap: done - thanks for the cross-check

pbiering:
Please add changelog entry related to changed option

pbiering (approved):
ok

pbiering (approved):
thank you

PR #1678, opened 2025-01-14
Add support for Dovecot auth over network

HmBMvXXiSivMcLGFWoqc:


pbiering:
one should mention that credentials are passed in cleartext over network if selected

pbiering (commented):
ok

PR #1689, opened 2025-02-02
migrate oauth2 into upstream

pbiering:
taken from https://gitlab.mim-libre.fr/alphabet/radicale_oauth/-/blob/dev/oauth2/ with adjustments

supports https://github.com/Kozea/Radicale/discussions/1359

### DIFF
--- a/radicale/auth/oauth2.py
+++ b/radicale/auth/oauth2.py
@@ -61,9 +61,6 @@ class Auth(auth.BaseAuth):
             ):
                 return login
         except OSError as e:
-            raise RuntimeError(
-                "Failed to authenticate against oauth server %r: %s"
-                % (self._endpoint, e)
-            ) from e
-        logger.warning("User %s failed to authenticate" % (str(login)))
+            logger.critical("Failed to authenticate against OAuth2 server %s: %s" % (self._endpoint, e))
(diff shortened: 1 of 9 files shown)


radicale/auth/oauth2.py
```diff
@@ -0,0 +1,66 @@
...
+    def _login(self, login, password):
+        """Validate credentials.
+        Sends login credentials to oauth token endpoint and checks that a token is returned
+        """
+        try:
+            # authenticate to authentication endpoint and return login if ok, else ""
+            req_params = {
+                "username": login,
+                "password": password,
+                "grant_type": "password",
```
benbucksch, line 51: This password grant type is rarely enabled, so this is weil usually not work.

Instead, the app sends the user to an auth page in the browser,  which returns an authCode, then the all uses the OAuth server to change that auth code into an access token and refresh token,  and then the app passed the access token to the app server (radicale), and the app server only needs to check whether the access token is valid.

pbiering: I would have assumed this already, see my comments in the discussions...if one really want to have OAuth2 inside radicale and not terminating on reverse proxy in front of, further code contribution is required (incl. all the error handling).

PR(s) are welcome!

PR #1742, opened 2025-03-24
Update ldap.py

BastelBaus:
Bugfix, user_entry['attributes'][self._ldap_user_attr] is already the string so user_entry['attributes'][self._ldap_user_attr][0] would give only the first character and not the full user attribute


BastelBaus:
P.S. Hope this is not double. I tried it already yesterday but it is vanished. I am not so experienced in git.

PR #1762, opened 2025-04-19
Add LDAP start_tls support

imgajus:
I've added support for LDAP auth start_tls encryption. Tested with Wireshark and Authentik LDAP outpost to confirm that the encryption functions correctly and added checks to prevent sending credentials through cleartext for start_tls. I think similar checks should be introduced for SSL too but I have no way of testing those currently.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1045,13 +1045,13 @@ Default: (unset)
 
 _(>= 3.3.0)_
 
-Use ssl on the ldap connection (soon to be deprecated, use ldap_encryption instead)
+Use ssl on the ldap connection (soon to be deprecated, use ldap_security instead)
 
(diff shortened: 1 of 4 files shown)


pbiering:
can you please rename `ldap_encryption` with `ldap_security` to be aligned with already existing `imap_security` which does the same for IMAP. Also for alignment rename `start_tls` with `starttls`. Thank you!

imgajus:
> Can you please rename `ldap_encryption` with `ldap_security` to be aligned with already existing `imap_security` which does the same for IMAP. Also for alignment rename `start_tls` with `starttls` and `ssl` with `tls`. Thank you!

Done!

pbiering (approved):
looks good now

pbiering:
thank you!

Issue #1626, opened 2024-11-21
python-dateutil seems to be unused


lahwaacz:
Searching for `dateutil` in this repository yields only three matches:

https://github.com/search?q=repo%3AKozea%2FRadicale%20dateutil&type=code

One in `pyproject.toml`, another one in `setup.py.legacy`, and the last in `radicale/storage/__init__.py` where it is only used as part of the `CACHE_VERSION` bytestring.

pbiering:
Thank you for reporting, looks like a legacy reference, nowadays all is handled by _vobject_

PR #1769, opened 2025-04-26
Remove obsolete dateutil reference

lahwaacz:
The dependency was removed, see https://github.com/Kozea/Radicale/issues/1626


pbiering (approved):
make sense, thank you!

PR #1775, opened 2025-04-30
Fix some incorrect grammar in the log

fiwswe:


fiwswe:
The original log text was ambiguous due to the incorrect English grammar. It should be more clear now.

pbiering (approved):
ok, thank you!

PR #1787, opened 2025-05-25
Docker publish workflow: Add architecture linux/arm64

kalsi-avneet:
From the listed future tasks in https://github.com/Kozea/Radicale/discussions/1745#discussioncomment-13004594

> Add support for other architectures (e.g. arm64)

This PR adds support for arm64. Other architectures can be added later if needed.


pbiering (approved):
ok

PR #1791, opened 2025-05-30
Implementing the evalutaion of indirect group membership.

petervarkoly:
Now member or uniqueMember can also be used for the group membership.

### DIFF
--- a/config
+++ b/config
@@ -92,10 +92,13 @@
 # Path of the file containing password of the reader DN
 #ldap_secret_file = /run/secrets/ldap_password
 
-# The attribute to read the group memberships. This can be memberOf from the user's LDAP entry. member or uniqueMember can also be used. In this case an additional ldap search will be executed to find the groups where the user is member of.
+# The attribute in user entry to read the group memberships.
 #ldap_groups_attribute =
 
-# The base dn to find the groups. Will be used only if ldap_groups_attribute is member or uniqueMember. If not given ldap_base will be used.
+# The attribute in group entries to read the group memberships.
+#ldap_group_members_attribute =
+
+# The base dn to find the groups. Necessary only if ldap_group_members_attribute is defined and other then ldap_base.
 #ldap_groups_base =
 
 # The filter to find the DN of the user. This filter must contain a python-style placeholder for the login
--- a/config
+++ b/config
@@ -92,8 +92,11 @@
 # Path of the file containing password of the reader DN
 #ldap_secret_file = /run/secrets/ldap_password
 
-# the attribute to read the group memberships from in the user's LDAP entry (default: not set)
-#ldap_groups_attribute = memberOf
+# The attribute to read the group memberships. This can be memberOf from the user's LDAP entry. member or uniqueMember can also be used. In this case an additional ldap search will be executed to find the groups where the user is member of.
+#ldap_groups_attribute =
+
+# The base dn to find the groups. Will be used only if ldap_groups_attribute is member or uniqueMember. If not given ldap_base will be used.
+#ldap_groups_base =
 
 # The filter to find the DN of the user. This filter must contain a python-style placeholder for the login
 #ldap_filter = (&(objectClass=person)(uid={0}))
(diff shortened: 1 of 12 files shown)


radicale/auth/ldap.py
```diff
@@ -130,8 +134,8 @@ def __init__(self, configuration: config.Configuration) -> None:
             else:
                 logger.info("auth.ldap_ssl_ca_file     : (not provided)")
         """Extend attributes to to be returned in the user query"""
-        if self._ldap_groups_attr:
-            self._ldap_attributes.append(self._ldap_groups_attr)
+        if self._ldap_groups_attr == "memberOf":
+            self._ldap_attributes.append("memberOf")
```
marschap, line 138: Sorry, this is not acceptable.
In order to fix "perceived" hard-coded values (which in reality are configurable), you now do hard-coding.
This breaks all setups that do not have ActiveDirectory's `memberof` attribute as an indicator of the user side.
Among them: 
* all setups using Novell eDirectoy's `groupMemberShip`
  There may be other directory servers using different schemes, but Novell eDirectory I know for sure.
  OpenLDAP even allows using a freely defined attribute for the user-side of groups (by using a modules)
  * all setup that use a non-DN valued attribute (like .e.g mine)

config
```diff
@@ -92,10 +92,13 @@
 # Path of the file containing password of the reader DN
 #ldap_secret_file = /run/secrets/ldap_password
 
-# The attribute to read the group memberships. This can be memberOf from the user's LDAP entry. member or uniqueMember can also be used. In this case an additional ldap search will be executed to find the groups where the user is member of.
+# The attribute in user entry to read the group memberships.
```
marschap, line 95: Please revert to the original comment before your commit (no unnecessary churn)

radicale/auth/ldap.py
```diff
@@ -275,12 +276,12 @@ def _login3(self, login: str, password: str) -> str:
             tmp: list[str] = []
             gdns: list[str] = []
             """Let's collect the groups of the user."""
-            if self._ldap_groups_attr == "memberOf":
-                gdns = user_entry['attributes']['memberOf']
-            elif self._ldap_groups_attr == "member" or self._ldap_groups_attr == "uniqueMember":
+            if self._ldap_groups_attr:
+                gdns = user_entry['attributes'][self._ldap_groups_attr]
+            elif self._ldap_group_members_attr:
                 conn.search(
```
marschap, line 282: You may want to consider doing this search as the same user having done the user search, and not as the user found.
I am positively aware of LDAP setups where users are not allowed to search for anything apart from parts of their own entry. In such a case this search - when attempted as the user found (i.e. after checking the password using the bind() as user - will not yield results / fail.

radicale/config.py
marschap, line 1: sorry, another oversight from the review of your 1st commit:
config.py: `ldap_groups_base_dn`   vs ldap.py `ldap_groups_base` ??
For consistency with the other variables, please change the name in config.py to `ldap_groups_base`.

marschap (changes requested):
Hi @petervarkoly,
please look at the comments in the review.
As it is hard to review two commits which build onto each other in one MR/PR,
I strongly recommend, to 
* split the PR into 3 PRs, each of them being one commit with the abslute minimal changes that logically belong together
    1. one for the changes around the LDAP groups
    2. one for the `conn.bind(read:_Server_info=TRUE)` thing 
      and properly assign it to the issue
    3. one for the space changes...
      They are cleanups that neither belong to the ones above
* test before you submit the respective PRs
   The `ldap_group_base_dn` vs. `ldap_group_base` discrepancy should have been noticed.

Please note: I only ran my radicale setup with your changes in `config.py` and in `auth/ldap.py` (successfully after renaming the `ldap_group_base` on `config.py). 
I did not perform functionality test on the new group-based config..
    3. one for the space changes...
      they do neither belong to 1. or 2.

pbiering:
either fix tests or cancel

marschap:
@petervarkoly You may want to have a look at the branch ldap.group-search in my fork marschap/Radicale.
I tested it with 
* ldap_groups_atrribute denoting
    * group valued attributes in the user
    * non-group attributes in the user (this is a supported scenario ;-))
 * searching for groups using te separate search
 
 I deliberately did not create a PR to let you the authorship.(as you did most of the work).
 It includes the try: ... except brace around the group query to cater for search failures.

pbiering:
superseeded by https://github.com/Kozea/Radicale/pull/1861

Issue #1789, opened 2025-05-26
ldap auth will grab the whole schema needlessly, every time


alien999999999:
I looked with tcpdump and every time a ldap thing happened, the whole schema was queried, which gives big load; I would prefer configurable ldap attribute names or something if needed.

pbiering:
@petervarkoly / @marschap - can you please take a look into?

petervarkoly:
Only certain attributes are required for all searches. 
Do you use ldap or ldap3?
Can you evaluate from the tcpdump by which step the whole schema will be required?

petervarkoly:
If it is ldap3 I have found the problem:
conn.bind() have to be called with read_server_info=False. Default is True.

pbiering:
any update here?

marschap:
IMHO this should be closed:  #1792, which already is in master, is the fix for the ldap3 part.

PS: I'd have closed it myself, but I lack the permissions.

pbiering:
closing now

PR #1792, opened 2025-05-31
Fix binding ldap server.

petervarkoly:
Do not read server info by binding to avoid needless network trafic.


petervarkoly:
Fixes issue https://github.com/Kozea/Radicale/issues/1789

Issue #1812, opened 2025-06-27
Radicale search is returning multiple events instead of one


lbt:
Using Radicale after 3.1.9 returns multiple events for a search

python-caldav 1.6.0 fails and I've created a minimal test case
https://gist.github.com/lbt/01b5932d80bc1c87eccd0695fe4189f8

This currently returns 11 events for a recurring event.

git bisect shows the problem appears to have been introduced in 513e04e63

commit by @metallerok 

python-caldav now has a workaround so 2.0.1 works again:
https://github.com/python-caldav/caldav/issues/439
https://github.com/python-caldav/caldav/commit/6f0bd41821ff2a19dc81b1eeb5d1b8e864cda0e0

pbiering:
this was introduced by following PR: https://github.com/Kozea/Radicale/pull/1293 

 @lbt : is python-caldav specifying the `expand` property during REPORT request or @metallerok is this a bug on server side?

metallerok:
The provided test example does not specify what event it was called for. Could you specify the event so that the error can be reproduced more accurately?

lbt:
@pbiering caldav does specify expand in 1.6.0 but not in 2.0.1 (due to their workaround)

Running against 3.5.4

I added this to the test case:

```
    import logging
    import http.client
    LOGGER = logging.getLogger(__name__)
    logging.basicConfig(level=logging.DEBUG)
    http.client.HTTPConnection.debuglevel = 5
```

1.6.0 caldav:

```
DEBUG:urllib3.connectionpool:Resetting dropped connection: group.dgreaves.com
send: b'REPORT /home/df8ffba2-9c13-11ed-b6e9-1912c7a834fa/ HTTP/1.1\r\nHost: group.dgreaves.com:5232\r\nUser-Agent: python-caldav/1.6.0\r\nAccept-Encoding: gzip, deflate\r\nAccept: text/xml, text/calendar\r\nConnection: keep-alive\r\nContent-Type: application/xml; charset="utf-8"\r\nDepth: 1\r\nContent-Length: 420\r\nAuthorization: XXXXX==\r\n\r\n'
[...]

PR #1815, opened 2025-07-01
bugfix and tests for issue #1812

lbt:
A report should include:
* An expanded event with an expand start inside a recurring event
* An expanded event with an expand start and end inside a recurring event


pbiering:
@metallerok - good start, but potentially it would make more sense to filter each vevent "candidate" before adding, means shift the filter in front of

```
            if not is_component_filled:
                vevent_component.vevent = vevent
                is_component_filled = True
            else:
                vevent_component.add(vevent)
```

https://github.com/Kozea/Radicale/blob/1c97345fc90d185b1d8a3ce1a1a78210f26aa934/radicale/app/report.py#L374C1-L378C45

This at least lowers the risk of an out-of-memory issue. Code speed should be the same but the intermediate memory footprint is smaller imho.

metallerok:
> @metallerok - good start, but potentially it would make more sense to filter each vevent "candidate" before adding, means shift the filter in front of
> 
> ```
>             if not is_component_filled:
>                 vevent_component.vevent = vevent
>                 is_component_filled = True
>             else:
>                 vevent_component.add(vevent)
> ```
> 
> https://github.com/Kozea/Radicale/blob/1c97345fc90d185b1d8a3ce1a1a78210f26aa934/radicale/app/report.py#L374C1-L378C45
> 
> This at least lowers the risk of an out-of-memory issue. Code speed should be the same but the intermediate memory footprint is smaller imho.

This block of code is inside the loop of events received from the rrule and at the beginning of this loop there is filtering of events

https://github.com/Kozea/Radicale/pull/1815/files#diff-9ef2a477015c09ae208dd12ffde38735ce8c7e0f5b84cffc395b45d7cff7dcdbR375-R383

The filtering that occurs outside the loop is the filtering of the original events from which we received extensions. We also need to add processing of exdates and entire day events

pbiering:
@metallerok - looks my initial implementation proposal was incomplete :-(

case "no expand in report request" -> keep flow as it is

case "expand in report request":
- exclude timerange filters during initial filter
- during expanding, apply remembered timerange filters on each expanded vevent before adding to the list to keep memory footprint low

pbiering:
@lbt  - please fix all issues reported by "lint": https://github.com/Kozea/Radicale/actions/runs/16100018699/job/45427946048?pr=1815

Best by a local "tox" run before next commits: https://github.com/Kozea/Radicale/wiki/Development-Testing-Release

pbiering:
@lbt + @metallerok : ready to merge?

pbiering:
> @lbt + @metallerok : ready to merge?

Further todos:

- radicale/tests/test_expand.py 
  - extend copyright section, if applicable
- radicale/app/report.py
  - extend copyright section, if applicable
  - add DoS protection for "expand" as for "freebusy", see https://github.com/Kozea/Radicale/blob/18710d9a272b64d020e2c5c284f3b7d535a96235/radicale/app/report.py#L132
    - amount of expanded entries limit reached (next new configurable "max_expand_occurence" option in "reporting"), log details of "expand" in error log, respond with " 400 Bad Request"

pbiering (approved):
ok, thank you!

metallerok:
@lbt I see the changes to return an empty response if we didn't find any events after filtering haven't been included. This probably needs to be added.
```python
    # Rebuild component
    if not filtered_vevents:
        element.text = ""
        return element
    else:
        vevent_component.vevent_list = filtered_vevents
        logger.debug("lbt: vevent_component %s", vevent_component)

    element.text = vevent_component.serialize()

    return element
```

lbt:
> @lbt I see the changes to return an empty response if we didn't find any events after filtering haven't been included. This probably needs to be added.

ah, my bad. I removed the extra logging in the rebase and I must have messed up in one of the conflicts. I've added it but I'll also redo the rebase and double check.

lbt:
I'm also looking at
```recurrences = rruleset.between(rstart, end, inc=True)```
and then later, for each entry there is a time-range filter check. This feels redundant but there's a utc conversion so maybe it has some use?

I'm looking at the DDOS part now. It seems reasonable and I notice rruleset.between takes count which could help too.
I'll see what I can do.

lbt:
still need to add some tests for maxfreebusy config

lbt:
I found an error in how vevents were counted. Probably not important as the main risk comes from a huge rrules expansion.

pbiering:
> I found an error in how vevents were counted. Probably not important as the main risk comes from a huge rrules expansion.

would you push a fix before I should merge?

lbt:
> > 
> 
> would you push a fix before I should merge?

ah, sorry, I meant I found and fixed because I added the test  case :)

pbiering:
> > > 
> > 
> > 
> > would you push a fix before I should merge?
> 
> ah, sorry, I meant I found and fixed because I added the test case :)

Ok, in this case I will merge now.

pbiering:
fixing https://github.com/Kozea/Radicale/issues/1812

PR #1822, opened 2025-07-17
Fix: Exclude empty calendars from the result if no matching events were found after applying filters

metallerok:
(#1812)
Exclude empty calendars from the result if no matching events were found after applying filters.
Excluding elements that do not match the comp-filter.


metallerok:
(#1812)
When extracting the time-range filter during expand processing, the entire comp-filter was excluded, which caused component filtering to fail. As a result, components that were not explicitly requested or should have been excluded based on filter parameters could appear in the output. Now, only the time-range is extracted from the filters without affecting the rest of the filtering process.

PR #1843, opened 2025-08-12
Fix - Storage hook only adds path to item in PUT

BlueRains:
Fix of issue #1842 - Storage hook only adds path to item in PUT.

- Added path to `self._storage.acquire_lock()` in all files that were missing it.
- Added tests to check the paths given to the hook are correct


pbiering:
Hmm, looks like timed collision, implemented the same (and even more): https://github.com/Kozea/Radicale/pull/1843

pbiering:
I will close this PR, but thank you anyhow. If you like you can improve a little bit my test code, your code with function looking smarter somehow...mine is more "spaghetti" like but covering also "request" and "to_path".

PR #1857, opened 2025-08-29
Fix broken start when UID does not exist

### LINKED ISSUE #1856: Since 3.5.5, unable to run rootless (`getpwuid` change?)
Hi,

Reporting here for users of the Docker image at https://github.com/tomsquest/docker-radicale/issues/200

Users reported permission issues that since `3.5.5`, for which I have no clue at all (no docker rootless here).
The issue may be related to #1830.

**TL;DR**
- ✅ OK 3.5.4 
- ❌ KO 3.5.5
- Error reported is about `getpwuid` for which changes where made, see #1830 

If you have any idea, any way to check what is going on, 👍

davidfdezalcoba:
Fixes #1856


pbiering (approved):
make sense

Issue #1859, opened 2025-08-30
send rip= to dovecot auth


jmberg:
For more advanced authentication setups, it can be useful to have the remote IP, e.g. to disallow using login passwords for remote connections (forcing service passwords).

I  tried to look at this, but it doesn't seem like the IP address is passed to authentication today, and at least right now I wasn't sure I understood the plugin system well.

pbiering:
implemented by https://github.com/Kozea/Radicale/pull/1860

PR #1860, opened 2025-08-30
auth: dovecot: pass rip= to auth server

jmberg:
If known, let the auth server know where the client came from, using (in this order) forwarded-for or remote-addr.

Addresses #1859.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Fix: broken start when UID does not exist (potential container startup case)
 * Improve: user/group retrievement for running service and directories
 * Extend/Improve: [auth] ldap: group membership lookup
+* Add: option [auth] dovecot_rip_x_remote_addr
 
 ## 3.5.5
 * Improve: [auth] ldap: do not read server info by bind to avoid needless network traffic
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1187,6 +1187,26 @@ Port of via network exposed dovecot socket
 
 Default: `12345`
 
+##### dovecot_rip_x_remote_addr
+
+_(>= 3.5.6)_
+
+Use the `X-Remote-Addr` value for the remote IP (rip) parameter in the
+dovecot authentication protocol.
+
+If set, Radicale must be running behind a proxy that you control and
+that sets/overwrites the `X-Remote-Addr` header (doesn't pass it) so
+that the value passed to dovecot is reliable. For example, for nginx,
+add
+
+```
+    proxy_set_header  X-Remote-Addr $remote_addr;
+```
+
+to the configuration sample.
+
+Default: `False`
+
 ##### imap_host
 
 _(>= 3.4.1)_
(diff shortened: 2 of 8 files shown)


pbiering (approved):
ok

pbiering:
"lint" workflow is unhappy, please fix and best verify with a local "tox" run before updating the PR.

See also https://github.com/Kozea/Radicale/wiki/Development-Testing-Release

jmberg:
Sorry, hadn't seen how to run tests (and I was sure you'd comment on e.g. the extra data class or other things ;-) )

pbiering:
@jmberg : `HTTP_X_FORWARDED_FOR` can contain more than one IP address (comma `,` separated) and cannot be trusted 100% because can be set up upstream proxy or even client.

See also: https://en.wikipedia.org/wiki/X-Forwarded-For

Dou you really want to use it for `rip` in dovecot auth, if provided?

jmberg:
Fundamentally, I think yes, particularly in my setup. Unless I'm misunderstanding things? I have radicale working behind a proxy that deals with https etc. (nginx, mostly like the same config), and I do want the authentication to know where the connection was originally from, otherwise it's always just ::1 or 127.0.0.1.

The multiple entries is an issue though, I thought you were using it already somewhere but I see it's just for logging right now.

Not sure how to square that - dovecot by default also only uses it for logging I believe, so maybe I can just document that the value is passed as-is (I need to quote/escape or at least remove tabs/newlines so it can't result in protocol injection even when there's no upstream proxy or it's not overriding it) and must be set by an administrator-controlled proxy to be reliable?

OTOH even that is stupid, that would mean if you don't even want a proxy but want to use rip=, you'd still need a proxy since I made it prefer X-Forwarded-For over REMOTE_ADDR, which means it can never be reliable ...

So I guess the right thing to do would be to use REMOTE_ADDR by default, make it configurable for X-Forwarded-For, and explain in the docs for that configuration that it _must_ be paired with a proxy that always _sets_ the value. Then maybe it could be argued I don't even need the escaping, but it's still different severity (wrong rip= vs. possibly authentication injection/bypass) so still better have that and strip any whitespace.

Let me see how to make things configurable.

pbiering:
I would also prefer by default only using REMOTE_ADDR and optional use e.g. X_REMOTE_ADDR filled by reverse proxy config with the REMOTE_ADDR it's seen. A 3rd option would be using the last entry in the ',' separated list (and hoping, all reverse proxies behave the same and append itself without reordering)  X-Forwarded-For, it can be simulated very easily by a client.

```
curl -H "X-Forwarded-For: 1.2.3.4" ...
```
results in
```
 HTTP_X_FORWARDED_FOR:  1.2.3.4, 192.168.x.y
```

192.168.x.y is the client address connecting to reverse proxy.

jmberg:
But certainly we can also say X-Remote-Addr instead, and document that similarly. Was doing this before I saw your new comment.

pbiering:
@jmberg - would be great if you can adjust this PR according to review.

PR #1868, opened 2025-09-09
auth: clean up remote IP parameters/documentation

jmberg:
Make the remote IP parameter more generic and make it an enum determining the source instead of a boolean. Also fix the changelog entry.

Both as requested, I managed to miss those comments previously.


jmberg:
My apologies, I missed the prior comments for no good reason whatsoever.  Sorry about that.

PR #1877, opened 2025-09-19
LDAP auth: move evaluation of quirk for Authentik where it belongs

marschap:
The evaluation of the quirk for the Authentik LDAP server changes the behaviour of Python's `ldap3` module, and that module only.
Evaluating the quirk in `__init__` which is used for both, `ldap` and `ldap3` is thus wrong, and may lead to errors when this setting is used together with the `ldap` module.


pbiering:
superseeded by https://github.com/Kozea/Radicale/pull/1884

PR #1885, opened 2025-09-25
LDAP auth: move evaluation of quirk for Authentik where it belongs

marschap:
The evaluation of the quirk for the Authentik LDAP server changes the behaviour of Python's `ldap3` module, and that module only.
Evaluating the quirk in `__init__` which is used for both, `ldap` and `ldap3` is thus wrong, and may lead to errors when this setting is used together with the `ldap` module.

This should address the issues with the previous PR.

(Sorry that it became a new PR - I accidentially deleted the branch of the old one)


pbiering:
@marschap  - can you please rebase, see https://github.com/Kozea/Radicale/pull/1884

pbiering (approved):
thank you!

PR #1887, opened 2025-09-28
Ldap auth overhaul

marschap:
Hi Peter,

this PR foremost brings LDAP encryption (tls / starttls) to radicale's `ldap` authentication module when used with Python's `ldap` module.
Until now, any LDAP encryption setting in the config was silently ignored when using the 'ldap' module instead of the 'ldap3' module, which I consider a big secuerity hole.

In addition, it
* prepares for the removal of `ldap_use_ssl`
* gets rid of the helper property `_use_encryption`
* infers `ldap_security=tls` when `ldap_uri` starts with `ldaps://`
* warns on conflicting settings of `ldap_use_ssl` and `ldap_security`
* makes Radicale fail on illegal values for the config settings `ldap_security` and `ldap_ssl_verify_mode`
* lowers `ldap_ssl_verify_mode` to NONE when `ldap_uri` starts with `ldapi://`

Please consider pulling into master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 3.5.8.dev
+* Extend [auth]: re-factor & overhaul LDPA autrhentication, especially for Python's ldap module
 
 ## 3.5.7
 * Extend: [auth] dovecot: add support for version >= 2.4
--- a/radicale/auth/ldap.py
+++ b/radicale/auth/ldap.py
@@ -119,6 +119,9 @@ class Auth(auth.BaseAuth):
         if self._ldap_uri.lower().startswith("ldaps://") and self._ldap_security not in ("tls", "starttls"):
             logger.info("Inferring 'ldap_security' = tls from 'ldap_uri' starting with 'ldaps://'")
             self._ldap_security = "tls"
+        if self._ldap_uri.lower().startswith("ldapi://") and self._ldap_ssl_verify_mode != "NONE":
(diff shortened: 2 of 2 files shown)


marschap:
Shall I also go the final step and add a commit for getting rid of `ldap_use_ssl` for good`?
Of course this can also be done in a separate PR at a later time.

pbiering:
@marschap - can you please rebase (3.5.8.dev) and add also related line in changelog?

pbiering:
> Shall I also go the final step and add a commit for getting rid of `ldap_use_ssl` for good`? Of course this can also be done in a separate PR at a later time.

in a minor release a known option should not disappear...would it be possible to create already now a code extension that automatically from Radicale >= 3.6.0 the option will be still detected but start will be prohibited then instead of displaying a warning as of now - then also in documentation this can be already mentioned that only supported < 3.6.0.

marschap:
PR rebaeed and extended with CHANGELOG.md updates (and adding my Copyright to radicale/auth/ldap.py)

marschap:
> > Shall I also go the final step and add a commit for getting rid of `ldap_use_ssl` for good`? Of course this can also be done in a separate PR at a later time.
> 
> in a minor release a known option should not disappear...would it be possible to create already now a code extension that automatically from Radicale >= 3.6.0 the option will be still detected but start will be prohibited then instead of displaying a warning as of now - then also in documentation this can be already mentioned that only supported < 3.6.0.

I concur that`ldap_use_ssl` shall be kept in the 3.5.x series.
I'll try to come up with a separate patch/PR over the next weekend.

pbiering (approved):
Thank you!

Issue #1897, opened 2025-10-15
Feature request: optional strict ETag / precondition enforcement on PUT


EdwinKM:
**Description:**
Currently, Radicale accepts PUT requests even if the client sends an outdated ETag or no If-Match header. This can lead to unexpected conflict resolution behavior, where the last client to write “wins”, regardless of the ETag.

In the examples i use "curl -i -u "user:password"   http://localhost:5232/user/9019b1ed-b1e9-324d-c127-ec06049312fb/d5d5af58-b4b6-4701-9d88-f4d3fefdb44c.vcf" to determine the ETag.
Observed the following scenarios with CardDAV clients:

**Conflict (Thunderbird wins — expected behavior)**

-   Ensure DAVx5 auto-sync is off
-   Start Radicale
-   Create user "user" in Thunderbird
-   Sync in Thunderbird
-   ETag = ed1a5b40dc08fb6c5dee177e40345712cbac95a779013eb7d94ee768d1fdb589
-   Sync in DAVx5
-   Verify "user" visible in Fossify
-   Stop Radicale
-   Rename in Fossify: "user" → "fossify"
-   Rename in Thunderbird: "user" → "thunderbird"
-   Stop Thunderbird and Fossify
-   Start Radicale
-   Start Thunderbird and sync
-   ETag = 7c9b87f81baac881637cd68c081fcd116d9c5b62071732b640319e2eaa8811ac
-   Sync in DAVx5

```
    [2025-10-15 19:29:49 +0000] [1/Thread-17 (process_request_thread)] [WARNING] Precondition failed on PUT request for '/user/9019b1ed-b1e9-324d-c127-ec06049312fb/d5d5af58-b4b6-4701-9d88-f4d3fefdb44c.vcf' (HTTP_IF_MATCH: "ed1a5b40dc08fb6c5dee177e40345712cbac95a779013eb7d94ee768d1fdb589", item has different etag: "7c9b87f81baac881637cd68c081fcd116d9c5b62071732b640319e2eaa8811ac")
    [2025-10-15 19:29:49 +0000] [1/Thread-17 (process_request_thread)] [INFO] PUT response status for '/user/9019b1ed-b1e9-324d-c127-ec06049312fb/d5d5af58-b4b6-4701-9d88-f4d3fefdb44c.vcf' in 0.497 seconds: 412 Precondition Failed
```
    

  _Result: Thunderbird wins, Fossify adopts the change — as expected._

**Conflict (Fossify should win)**

[...]

PR #1900, opened 2025-10-20
Additional etag tests

pbiering:


pbiering:
relates to https://github.com/Kozea/Radicale/issues/1897

PR #1951, opened 2026-01-03
fix misindented early return

### LINKED ISSUE #1917: Sending mail only, when removing attendee from event (not at generating event)
Hello all, 

first of all so much for this great type of software and all the efforts put in!
I loved to try out sending generated mail, when creating an event with specific attendees. 

I configured it like that: 

```
[hook]
type=email
smtp_server=xxx
smtp_port=465
smtp_security=tls
smtp_ssl_verify_mode=REQUIRED
smtp_username=xxx
smtp_password=xxx
from_email=xxx
```

Then I generated events with attendees, but I did not receive an e-mail notification (which was I thought I misconfigured my mail provider). 
However when I cancelled the test events, I received a mail, that the event was cancelled (so mail provider works). 

The question is now, why are mail notification working when deleting an event and not while creating one. 

Thank you!
Raemsna

samueltardieu:
The wrong indentation could cause the mail hook to bail early without any logging when the end time of the event being added/modified happened later than the current boundary being checked (1 full minute before the current time).

For example, this caused emails not to be sent in the following setup: Etar + DAVx⁵ on Android.

Fixes #1917

### DIFF
--- a/radicale/hook/email/__init__.py
+++ b/radicale/hook/email/__init__.py
@@ -988,7 +988,7 @@ class Hook(BaseHook):
                 if event_end < (now - timedelta(minutes=1)):
                     logger.warning("Event end time is in the past, skipping notification for event: %s",
                                    email_event_event.uid)
-                return
+                    return
 
             if not previous_item_str:
                 # Dealing with a completely new event, no previous content to compare against.


pbiering (approved):
ok

samueltardieu:
I've added "Fixes #1917" in the description, I noticed the issue afterwards.

Issue #1952, opened 2026-01-05
Use of bcrypt fails because passlib is no longer maintained


danielfdickinson:
FYI, passlib is no longer maintained and this causes issues where bcrypt errors (complains password needs to be longer than 72 bytes, even with passwords of e.g. 16 bytes).

Ansible has a patch, but other projects are considering dropping passlib.

https://forum.ansible.com/t/python-passlib-and-bcrypt-issue-and-patch/44640

https://github.com/pyca/bcrypt/issues/684#issuecomment-3474349636

https://stackoverflow.com/questions/79469355/error-attributeerror-problem-with-passlibbcrypt-module

pbiering:
This is related to https://github.com/Kozea/Radicale/issues/1896

pbiering:
There is a fork on github which looks active: https://github.com/notypecheck/passlib also having a new PyPI project name https://pypi.org/project/libpass/

pbiering:
Can you please test applying changes from https://github.com/Kozea/Radicale/pull/1953 whether it works at least with the fork?

danielfdickinson:
Thank you. Will do so, though not sure when yet (busy work week coming up).

pbiering:
will be fixed with 3.6.0 (3.5.11 will be skipped)

pbiering:
latest upstream contains now a check whether bcrypt version is supported by passlib(libpass) version - should be helpful hopefully.

Wiki also updated https://github.com/Kozea/Radicale/wiki#external-modules

PR #1953, opened 2026-01-05
replace passlib with libpass

pbiering:
replace passlib (no longer maintained since 2020) with libpass

Supports https://github.com/Kozea/Radicale/issues/1952


danielfdickinson:
I can confirm Radicale 3.5.11.dev from this branch works correctly.

I have tested bcrypt and md5 with this setup, and they 'just work'.

If you are curious, I have a test OpenWrt package which uses this branch (and adds python-libpass) at https://gitlab.com/dfd-web/firmware/openwrt-packages/-/tree/radicale3-libpass?ref_type=heads .

PR #1955, opened 2026-01-06
Passlib libpass check related to bcrypt

pbiering:
supports https://github.com/Kozea/Radicale/issues/1952 by adding check whether bcrypt version is supported by passlib/libpass version


radicale/utils.py
```diff
@@ -27,6 +27,8 @@
 from string import ascii_letters, digits, punctuation
 from typing import Callable, Sequence, Tuple, Type, TypeVar, Union
 
+from packaging.version import Version
```
danielfdickinson: @pbiering This adds a new external dependency on the packaging module (I found out with my OpenWrt build).

pbiering: thank you, new dependency added by 184a4491d84efaf2b2de42f26aacdd3dab7e0ddf

Wiki also extended: https://github.com/Kozea/Radicale/wiki#external-modules

PR #1950, opened 2026-01-01
Create docker compose file and modify documentation

kalsi-avneet:
Created a docker compose file and modified documentation for it. Users will be able to deploy Radicale easily via the compose file.

This was also requested in https://github.com/Kozea/Radicale/issues/1941

**Please note**:
For now, the latest release (`3.5.10`) is hardcoded in the compose file. After next release, I will create another PR to change it to `stable`. This tag will always point to the latest stable docker image version.

The CI file was modified to add the `stable` tag in https://github.com/Kozea/Radicale/pull/1944. This tag will appear when CI runs for the next release.

PR #1963, opened 2026-01-11
Docker compose file - change tag to "stable"

kalsi-avneet:
In https://github.com/Kozea/Radicale/pull/1950

> **Please note**:
> For now, the latest release (`3.5.10`) is hardcoded in the compose file. After next release, I will create another PR to change it to `stable`. This tag will always point to the latest stable docker image version.

This PR makes the above change

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,7 +1,7 @@
 name: Radicale
 services:
     radicale:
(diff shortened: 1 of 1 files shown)


pbiering (approved):
ok

PR #1968, opened 2026-01-21
Fix MOVE failing with URL-encoded Destination header

laurisvr:
MOVE requests fail with 502 Bad Gateway when the Destination header contains URL-encoded characters like `%40` for `@`.

I ran into this using email addresses as usernames (e.g. `user@example.com`). The client (Roundcube/Kolab) sends the Destination header as a relative URL with the `@` encoded:

```
Destination: /user%40example.com/calendar2/event.ics
```

Two issues in the original code:

1. The raw Destination wasn't being decoded before `urlparse()`
2. When Destination is a relative URL, `to_url.netloc` is empty, so `to_netloc_with_port` becomes just `:80` which never matches `server:80`

Fix: decode with `unquote()` first, and skip the netloc check for relative URLs.

Before:
```
Unsupported destination address: '/user%40example.com/calendar2/event.ics'
MOVE response status: 502 Bad Gateway
```

After:
```
MOVE response status: 201 Created
```

### DIFF
--- a/radicale/app/move.py
+++ b/radicale/app/move.py
@@ -22,7 +22,7 @@ import errno
 import posixpath
 import re
 from http import client
-from urllib.parse import urlparse
+from urllib.parse import unquote, urlparse
 
 from radicale import httputils, pathutils, storage, types
 from radicale.app.base import Access, ApplicationBase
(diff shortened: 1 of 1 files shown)


pbiering:
test cases added by https://github.com/Kozea/Radicale/pull/1969

PR #1973, opened 2026-01-23
Fix unclosed scandir iterator in path_to_filesystem

### LINKED ISSUE #1972: ResourceWarning: unclosed scandir iterator in pathutils.py
## Description

In `radicale/pathutils.py` line 290, `os.scandir()` is used in a generator expression without being properly closed, causing `ResourceWarning: unclosed scandir iterator` warnings.

## Current code

```python
if (os.path.lexists(safe_path) and
        part not in (e.name for e in os.scandir(safe_path_parent))):
```

## Suggested fix

```python
with os.scandir(safe_path_parent) as entries:
    if (os.path.lexists(safe_path) and
            part not in (e.name for e in entries)):
```

## Environment

- Radicale version: 3.5.10
- Python version: 3.14.2

## Steps to reproduce

Run pytest with warnings as errors:

```python
import warnings
warnings.filterwarnings("error")

# Use radicale...
```

The warning appears because `os.scandir()` returns an iterator that should be explicitly closed via context manager. While Python's garbage collector eventually closes it, this causes ResourceWarnings in environments that track unclosed resources.

---
🤖 *This issue was created with assistance from [Claude Code](https://claude.com/claude-code)*

tobixen:
## Summary
- Use `os.scandir()` as a context manager to properly close the iterator
- Add unit test that catches the ResourceWarning from unclosed iterators
- Fixes #1972

## Details

The `path_to_filesystem` function in `radicale/pathutils.py` uses `os.scandir()` in a generator expression without properly closing the iterator. When the iterator is garbage collected, Python emits a `ResourceWarning: unclosed scandir iterator`.

**Before:**
```python
if (os.path.lexists(safe_path) and
        part not in (e.name for e in os.scandir(safe_path_parent))):
    raise CollidingPathError(part)
```

**After:**
```python
if os.path.lexists(safe_path):
    with os.scandir(safe_path_parent) as entries:
        if part not in (e.name for e in entries):
            raise CollidingPathError(part)
```

## Test plan
- [x] Added unit test `test_scandir_iterator_closed` that verifies the iterator is properly closed
- [x] Test fails without the fix (ResourceWarning converted to error by pytest markers)
- [x] Test passes with the fix
- [x] All existing storage tests pass

### DIFF
--- a/radicale/pathutils.py
+++ b/radicale/pathutils.py
@@ -286,9 +286,10 @@ def path_to_filesystem(root: str, sane_path: str) -> str:
         safe_path = os.path.join(safe_path, part)
         # Check for conflicting files (e.g. case-insensitive file systems
         # or short names on Windows file systems)
-        if (os.path.lexists(safe_path) and
-                part not in (e.name for e in os.scandir(safe_path_parent))):
-            raise CollidingPathError(part)
+        if os.path.lexists(safe_path):
+            with os.scandir(safe_path_parent) as entries:
+                if part not in (e.name for e in entries):
+                    raise CollidingPathError(part)
     return safe_path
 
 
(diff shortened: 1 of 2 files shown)


tobixen:
I'm lazy nowadays, AI-generated pull request, but it seems like a simple and correct fix to me.

PR #1974, opened 2026-01-24
fix lighttpd example

schmijoe:
The lighttpd reverse proxy example in the documentation superfluously rewrites the path, resulting in warnings of the form `[WARNING] Called by reverse proxy, cannot remove base prefix '/radicale' from path: '/.web/css/icons/edit.svg' as not matching`.

This simplifies the configuration and removes the warnings.
Also setting the header `Script-Name` has no effect, so this is also removed.

### DIFF
--- a/contrib/lighttpd/radicale.conf
+++ b/contrib/lighttpd/radicale.conf
@@ -1,0 +1,6 @@
+server.modules += ( "mod_proxy" , "mod_setenv" )
+
+$HTTP["url"] =~ "^/radicale/" {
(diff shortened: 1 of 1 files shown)


pbiering:
can you also add example config to https://github.com/Kozea/Radicale/tree/master/contrib please into same PR?

schmijoe:
I've added a simple example config. (Identical to the one in the documentation)
Maybe it would also be nice to also provide an example of authentication via lighttpd, but I don't have much time for testing at the moment.

Issue #1966, opened 2026-01-19
Comparison error when sorting events based on a/the date/datetime of the event


lebouquetin:
We use `Radicale` with two clients: *Thunderbird* and [Open-Calendar](https://github.com/algoo/open-calendar/)

In some case, we got a `HTTP 500` with the following traceback:

```
2025-12-08 08:37:20 ERROR [radicale:136][uWSGIWorker2Core0] An exception occurred during REPORT request on '/agenda/workspace/30/': can't compare datetime.datetime to datetime.date
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/__init__.py", line 134, in __call__
    status_text, headers, answers = self._handle_request(environ)
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/__init__.py", line 366, in _handle_request
    status, headers, answer = function(
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/report.py", line 821, in do_REPORT
    status, xml_answer = xml_report(
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/report.py", line 313, in xml_report
    (expanded_element, n_vev) = _expand(
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/report.py", line 379, in _expand
    base_vevent, vevents_overridden = _split_overridden_vevents(vevent_component)
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/report.py", line 645, in _split_overridden_vevents
    vevent_recurrence, sorted(
TypeError: can't compare datetime.datetime to datetime.date
```
or later:

```
2026-01-19 13:38:23 ERROR [radicale:136][uWSGIWorker4Core0] An exception occurred during REPORT request on '/agenda/workspace/30/': can't compare datetime.datetime to datetime.date
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/__init__.py", line 134, in __call__
    status_text, headers, answers = self._handle_request(environ)
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/__init__.py", line 366, in _handle_request
    status, headers, answer = function(
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/report.py", line 821, in do_REPORT
    status, xml_answer = xml_report(
  File "/usr/local/lib/python3.9/dist-packages/radicale/app/report.py", line 313, in xml_report
    (expanded_element, n_vev) = _expand(
[...]

PR #1981, opened 2026-02-03
(#1966) overridden recurrence validation

metallerok:


pbiering:
@metallerok can you please add also test cases for that based on the reported issue?

metallerok:
Yes, i make this PR draft, because it doesn't have tests yet

PR #1983, opened 2026-02-08
Incorrect comma deleted and indentation corrected in predefined_colle…

cryinkfly:
I made a small mistake during the initial setup. Therefore, here's the correction. Otherwise, if the user uses it as a template, the container won't start.


pbiering (approved):
Thank you!

PR #1989, opened 2026-02-18
Refactor Web UI Javascript into separate files

maxberger:
This will simplify further maintainance. Instead of a single large file, there are now separate files for each models, api, utils, and constants. There are also separate files for all Scenes.

There are no functional changes in this patch.

### DIFF
--- a/integ_tests/README.md
+++ b/integ_tests/README.md
@@ -1,0 +1,28 @@
+# Integration tests
+
+These tests use Playwright to run the tests agains a full Radicale instance.
+
+## First time install
+
+You need to install playwright. The easiest way to do this is to create a .venv with playwright inside. You can do this by trying to run the integ tests and then entering the .venv
+
+Assuming you use uv to run/create your .venvs, note that the invocation will fail:
+
+``` lang=shell
+uv run --extra integ_test pytest integ_tests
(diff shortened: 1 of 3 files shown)


pbiering:
@maxberger - potenitally it makes sense to extend the tests for the internal web plugin by extending `radicale/tests/test_web.py`

maxberger:
I tested manually, but the point of adding automated tests made sense. Unfortunately the tests need to run inside a browser, so they require additional tooling.

I have created an extra integration_tests directory which is using playwright to run javascript integration tests. Right now there is a test for loading the index page, and one which actually tries to log-in. The second one will only work if the login javascript actually works.

pbiering (approved):
Thank you very much

pbiering:
> I tested manually, but the point of adding automated tests made sense. Unfortunately the tests need to run inside a browser, so they require additional tooling.

This can be overkill for now.

> I have created an extra integration_tests directory which is using playwright to run javascript integration tests. Right now there is a test for loading the index page, and one which actually tries to log-in. The second one will only work if the login javascript actually works.

That's great and enough for the moment!

pbiering:
@MatthewHana - fyi

PR #2002, opened 2026-02-27
Improve performance of `path_to_filesystem()`

flesueur:
When using radicale with a large set of accounts, the `PROPFIND` request at `/`, even with simple `owners_only` access rights, is very long (more than 100 seconds with 100% CPU on my test).

The bottleneck seems to be in the function `path_to_filesystem()` which is called for every folder (so for every account, when listing `/`), more precisely in the part in which it lists again this large folder.

This patch aims to check the same cases (short names on windows and case-insensitivity) with a different approach. In my test, the `PROPFIND` at `/` goes from 120 seconds to 3 seconds.

### DIFF
--- a/radicale/pathutils.py
+++ b/radicale/pathutils.py
@@ -285,9 +285,8 @@ def path_to_filesystem(root: str, sane_path: str) -> str:
         safe_path = os.path.join(safe_path, part)
         # Check for conflicting files (e.g. case-insensitive file systems
         # or short names on Windows file systems)
-        if (os.path.lexists(safe_path) and
-            not os.path.realpath(safe_path).endswith(part)):
-                raise CollidingPathError(part)
+        if (os.path.lexists(safe_path) and not os.path.realpath(safe_path).endswith(part)):
+            raise CollidingPathError(part)
     return safe_path
 
 


pbiering:
`flake8` is not happy:

flake8: commands[0]> flake8 .
./radicale/pathutils.py:285:9: F841 local variable 'safe_path_parent' is assigned to but never used
./radicale/pathutils.py:290:11: E128 continuation line under-indented for visual indent

-> please improve your PR, see also https://github.com/Kozea/Radicale/wiki/Development-Testing-Release

MatthewHana:
Haven't tested it yet but thanks for this @flesueur. I was running into this bottleneck when I was debugging another issue and was going to have a look into the performance degradation later- glad that now I don't have to.

pbiering:
> `flake8` is not happy:
> 
> flake8: commands[0]> flake8 . ./radicale/pathutils.py:285:9: F841 local variable 'safe_path_parent' is assigned to but never used ./radicale/pathutils.py:290:11: E128 continuation line under-indented for visual indent
> 
> -> please improve your PR, see also https://github.com/Kozea/Radicale/wiki/Development-Testing-Release

have fixed this now inline

flesueur:
Sorry I was offline during this weekend. Thanks for having cleaned this PR !

pbiering:
@flesueur - this change caused an exception in case softlinks are used to share collections between users:

```
Filesystem layout:

groups/ABOOK1-SHARED.vcf (file)
USER1/ABOOK1.vcf -> ../groups/ABOOK1-SHARED.vcf (softlink)

With some log:
root='/var/lib/radicale/collections/collection-root' sane_path='USER1/ABOOK1.vcf'
check part='USER1'
os.path.lexists=True os.path.realpath='/var/lib/radicale/collections/collection-root/USER1'
check part='ABOOK1.vcf'
os.path.lexists=True os.path.realpath='/var/lib/radicale/collections/collection-root/groups/ABOOK1-SHARED.vcf'
Exception: File name collision: 'ABOOK1.vcf'
```

Have applied this quick-fix now, please verify, potentially additional test cases are required and code can be improved to be faster related the condition checks:

```
--- a/radicale/pathutils.py
+++ b/radicale/pathutils.py
@@ -286,7 +286,7 @@ def path_to_filesystem(root: str, sane_path: str) -> str:
         safe_path = os.path.join(safe_path, part)
         # Check for conflicting files (e.g. case-insensitive file systems
         # or short names on Windows file systems)
-        if (os.path.lexists(safe_path) and not os.path.realpath(safe_path).endswith(part)):
+        if (os.path.lexists(safe_path) and not os.path.realpath(safe_path).endswith(part)) and not os.path.islink(safe_path):
             raise CollidingPathError(part)
     return safe_path
 ```

**Comment by flesueur** [comment]
This problem with symlinks seems logical, I didn't think about it.

Considering the performance, I don't see any problem with your fix.

However, concerning the check, I think there can be a problem. With your fix, a symlink will always validate, even if it is colliding (ie, a symlink with windows MSDOS names or a symlink on a case-insensitive filesystem). I'm not sure there are symlinks on Windows but there seems to be something like that and Python seems to handle them. However, I could not find a good solution : `realpath()` seems to be the only function to fully resolve filenames like "xxxxxx~1" on Windows but it also always follows symlinks.

Do you have any other idea on this topic ?

**Comment by pbiering** [comment]
At least I have merged now a test case for Unix-like filesystems. Anyhow, I have my doubts about the whole CollidingPathError at all. It's nowhere proper catched and imho only required in case of storing a file. But I've detected this during a PROPFIND request which is a read-only one and CollidingPathError makes no sense here at all. Only for "upload", "move", "create_collection" such check would imho make sense, so `path_to_filesystem` function looks like requiring an additional option with default "skip_colliding_check=False" and potentially a simulation option (PYTEST_RADICALE_*) set via environment to make this proper testable on non-Windows systems.

**Comment by flesueur** [comment]
Could it have some security implications on windows ? Confusing between the path radicale checks for access rights and the path that is accessed on the filesystem, if users trick the query with `~1` at the end of some path component ? I can't find the specific scenario but I think there may be some unexpected behavior. I can't test right now on windows unfortunately but I originally thought some part of this check was here for that.

**Comment by pbiering** [comment]
I will try to work on some test cases using Linux + vfat system. Track this now via https://github.com/Kozea/Radicale/issues/2081

**Comment by pbiering** [comment]
After adding a test case It was detected that your change broke collision detection for upcase vs. lowcase files. Will check how to improve that all.

**Comment by flesueur** [comment]
Since `realpath()` implementation of CPython is different on Windows and POSIX, Linux+VFAT may yield different results from Windows directly ? On a windows system with case-insensitivity I obtain the following results, which seem ok, `realpath()` gets the value of the existing file :

```
  Directory: C:\Users\User\Desktop\test


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         4/13/2026  12:21 AM              0 test.txt
-a----         4/13/2026  12:22 AM              0 Test2.txt


PS C:\Users\User\Desktop\test> python.exe
Python 3.13.12 (tags/v3.13.12:1cbe481, Feb  3 2026, 18:22:25) [MSC v.1944 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.realpath('test.txt')
'C:\\Users\\User\\Desktop\\test\\test.txt'
>>> os.path.realpath('test2.txt')
'C:\\Users\\User\\Desktop\\test\\Test2.txt'
```

pbiering:
For during startup detected non collision-free file systems this change is now reverted by https://github.com/Kozea/Radicale/pull/2087.

On collision-free file systems the collision-detection code is skipped completly now.

pbiering:
Major reason for reverting was that it broke support of symlinks (even on HFS+ or NTFS) found during PROPFIND.

PR #2006, opened 2026-03-01
Add UI for share by token

maxberger:
Initial implementation of sharing UI, allowing to create and remove shares by token


pbiering:
@maxberger can you please also add change for pyproject.toml and setup.py.legacy to change dedicated svg entries also to wildcard as already done for *.js?

maxberger:
> @maxberger can you please also add change for pyproject.toml and setup.py.legacy to change dedicated svg entries also to wildcard as already done for *.js?

done

PR #2010, opened 2026-03-02
Added Proper type hinting and type checking

maxberger:
This adds type hints and type checking in the javascript part of the webapp. Not all issues are addressed yet, but most of the warnings are now fixed.

This identified several bugs which were fixed on the side.


pbiering (approved):
ok

PR #2014, opened 2026-03-04
Misc improvements to UI

maxberger:
* Add github workflow which actually runs the integ tests
* DeleteCollectionScene: Fix JS linter errors
* Add confirmation on use of delete button
* Fix: Show only tokens in share by token list


pbiering (changes requested):
would it not better to merge the `integ_test.yml` into current `test.yml` to be able to use`needs` and start integration tests only after "test-ubuntu-python-newest" has successfully passed?

maxberger:
> would it not better to merge the `integ_test.yml` into current `test.yml` to be able to use`needs` and start integration tests only after "test-ubuntu-python-newest" has successfully passed?

done

pbiering:
@maxberger - is the "JS Type Check" independent from "integ-test" or should this be cascaded? If second, please add an additional dependency.

maxberger:
> @maxberger - is the "JS Type Check" independent from "integ-test" or should this be cascaded? If second, please add an additional dependency.

Both are independent. The js type check is more like a linter (do we have valid JavaScript?), the integ test tests everything including the server (when I click here does the right thing happen). 

I could even think about adding unit tests - but right now I think that would be too much.

PR #2025, opened 2026-03-15
Add support for managing incoming shares via map

maxberger:
This adds a new share button on the bottom right, which shows incoming shares via map; it let's the user enable/disable the share, and display it.

It uses a shown semantic instead of hidden, as this seems more intuitive to use.


pbiering:
> This adds a new share button on the bottom right, which shows incoming shares via map; it let's the user enable/disable the share, and display it.

Very nice!

PR #2032, opened 2026-03-20
Move CSP to config and update changelog

maxberger:


pbiering:
doc extended by b5e34fbe116b101667ac353fac40f08226b53fb6

config.py extension is not required, misunderstood way of how it is implemented.

maxberger:
Thanks, and sorry for the late reply!

PR #2036, opened 2026-03-22
Improve Authentication handling in webUI

maxberger:
* WebUI now properly recognizes pre-authentication, e.g. through X-Remote-User
* WebUI now uses Authentication header instead of putting the password in the URL

Unfortunately this does cause a message in the console since the authentication is checked on first load; that message can be safely ignored.

### DIFF
--- a/integ_tests/test_x_remote_user.py
+++ b/integ_tests/test_x_remote_user.py
@@ -153,11 +153,33 @@ def test_create_collection_works(
 
     create_collection(page, radicale_server)
 
-    # Verify that the new collection exists afterwards
-    # By default it's called "Untitled Collection" or similar?
-    # Let's check what create_collection does.
-    # It clicks .fabcontainer a[data-name="new"] and then #createcollectionscene button[data-name="submit"]
-
     expect(page.locator("article:not(.hidden)")).to_have_count(1)
-    # The title might be the HREF if no display name is set
     expect(page.locator("article:not(.hidden) .title")).to_be_visible()
+
+
+def test_download_works_with_remote_user(
+    context: BrowserContext, page: Page, radicale_server: str
+) -> None:
+    """Test downloading a collection with remote user."""
+    context.set_extra_http_headers({"X-Remote-User": "admin"})
+    page.goto(radicale_server)
+
+    # Wait for auto-login
+    expect(
+        page.locator(
+            '#logoutview span[data-name="user"]', has_text="admin's Collections"
+        )
+    ).to_be_visible()
+
+    create_collection(page, radicale_server)
+
+    # Start waiting for the download
+    with page.expect_download() as download_info:
+        # Perform the action that initiates download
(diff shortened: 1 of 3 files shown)


pbiering:
> WebUI now properly recognizes pre-authentication, e.g. through X-Remote-User

the original behavior is intended, reverse proxies have to exclude` /.web` from authentication requirement as the related example configurations are showing...same need to be applied (excluded) for upcoming `/.token` - can you please remove that part from the PR?

maxberger:
> > WebUI now properly recognizes pre-authentication, e.g. through X-Remote-User
> 
> the original behavior is intended, reverse proxies have to exclude` /.web` from authentication requirement as the related example configurations are showing...same need to be applied (excluded) for upcoming `/.token` - can you please remove that part from the PR?

I think there is some confusion here: The check is not on the .web/ directory itself. index.html, and all .JS in it is loaded as is, and as you said needs to be free from authentication.

What this does internally is: Before showing the login screen, it tries a propfind for the current users collection without credentials. If this one returns 401, we know we need to send all future requests with user/pass (previous behavior).  The only difference you see is the result of this call in the console.

However, if the call succeeds, e.g. because the user has valid credentials already (e.g. through x-remote-user or other methods, of course only if configured as auth), then this and all future calls are sent without username/password; they would not matter anyways, and the login screen can be skipped.

pbiering:
@maxberger  - the "JS Type Check" fails, is this a temporary glitch? 
https://github.com/Kozea/Radicale/actions/runs/23474116431/job/68309463903

maxberger:
> @maxberger  - the "JS Type Check" fails, is this a temporary glitch? 
> https://github.com/Kozea/Radicale/actions/runs/23474116431/job/68309463903

No, that is an issue. Looking....

maxberger:
> > @maxberger  - the "JS Type Check" fails, is this a temporary glitch?
> > https://github.com/Kozea/Radicale/actions/runs/23474116431/job/68309463903
> 
> No, that is an issue. Looking....

Breaking this down:

a) The check does not properly annotate the diff, so the current junit-like output does not work. I will need to fix the check, probably going to the boring output instead of the junit annotations. The good news is that the test failed.

b) the TSC compiler used by npx got updated to version 6 (previous was 5.9). Since the check always pulls the latest version, it pulled a new compiler, with new type checks! Unfortunately my local one was still on 5.9, so I did not see these issues. The good news is that they are not new -> So no new regressions. The bad news is that the warnings make sense (and will result in strange behavior in certain cases), so they need to be addressed.

Given that there is no immediate regression other than the read light, I would plan to address all of these properly (which means later this week) rather than immediately.

pbiering:
No hurry, release is postponed until Eastern, latest changes (PR pending) will cause some additional work for you...

maxberger:
Path proposed: https://github.com/Kozea/Radicale/pull/2042

PR #2052, opened 2026-03-28
remove legacy bday sharing

maxberger:


pbiering:
hmm, work collision, did the same...let see about whether there is a leftover...

PR #2053, opened 2026-03-28
Sharing remove legacy bday 2

pbiering:


pbiering:
@maxberger - leftovers found and covered with that PR...after merge I would assume all well done :-)

PR #2055, opened 2026-03-28
Run HTML/CSS validator in Tox/Github

maxberger:
Posting this separately, because I am again not sure if we want to fully enforce this. Although it is a good idea, and it allowed me to catch some errors in the HTML/CSS.

Requires PR #2054 before the tests succeed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -206,6 +206,20 @@ jobs:
       - name: Lint
         run: tox -c pyproject.toml -e flake8,mypy,isort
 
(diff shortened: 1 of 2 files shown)


pbiering:
Rerun of failed test triggered, still failed, does it need rebase and repush?

maxberger:
> Rerun of failed test triggered, still failed, does it need rebase and repush?

Apparently it does. Did exactly that, Let's see what happens.

pbiering (approved):
thank you

PR #2076, opened 2026-04-09
Sharing token permission default rp

pbiering:
- change default sharing permissions of token to "rp"
- fix bug in permission handling related to properties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Fix: share address book collection as birthday calendar not working on non-DEBUG level
 * Extension: share accept now PATH and USER for matching email addresses as well
 * Adjustment: replace logging/trace_on_debug by new log level "trace"
+* Adjustment: sharing/token: adjust default permissions to "rp"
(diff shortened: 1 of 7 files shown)


pbiering:
@maxberger - please check whether your WebUI has own default permissions defined, tokens should have "rp" by default for security reasons.

maxberger:
As a quickfix, I added the "p" permissions for token shares:
https://github.com/Kozea/Radicale/pull/2078

this would fix p, but still let the user chose between r and rw; This is good for the near future.

However, long term this is inconsistent. We should either:
a) Fix both r and p  for tokens
b) let the user chose between r and rw and between p and P

Which solution would you prefer?

pbiering:
Thank you for the quick fix

for shares and token owner should active have to set
- w (for write)
- P for properties overlay enabled

Optional:
- p for properties overlay disabled in case of globally or per rights enabled (can be detected by current propfind privileges indirectly)

Potentially this need some online help as it is somehow an expert option...

maxberger:
Added PR: https://github.com/Kozea/Radicale/pull/2082

PR #2078, opened 2026-04-10
UI: Set 'p' permission for tokens

maxberger:


pbiering (approved):
ok as a quick fix for now

PR #2079, opened 2026-04-10
fix(auth): prevent login variable overwrite during failed cache cleanup

jbsky:
In the failed login cache cleanup loop, the tuple unpacking was using `login` as the local variable name, silently overwriting the current user's login with the one retrieved from the expired cache entry.

This caused subsequent backend authentication (e.g. IMAP) to be attempted with the wrong username, resulting in spurious auth failures for legitimate users whenever an expired failed-cache entry happened to be present at the same time.

Rename the loop variable to `login_expired` to avoid shadowing the outer `login` variable, and fix the associated debug log to reference `login_expired` instead of the previously incorrect `login_cache`.

Fixes: intermittent authentication failures when _cache_failed contains expired entries from previous failed login attempts by other users.


pbiering (approved):
Good catch, thank you very much!

Issue #2081, opened 2026-04-11
pathutils/path_to_filesystem() need review


pbiering:
related to https://github.com/Kozea/Radicale/pull/2002, further review is required

TODO: some test cases using Linux + vfat system

pbiering:
Test cases extended, tested on Linux with XFS,VFAT,NTFS,HFS+ and native Windows+MacOS by workflow.

PR #2087, opened 2026-04-14
Issue 2081 / path_to_filesystem

pbiering:
* Fix: broken storage/mtime granularity detection on vfat
* Improve: `path_to_filesystem()` by pre-detection of collision-free file system
* Adjustment: MKCOL/MKCALENDAR return now CONFLICT instead of BADREQUEST of file name collision

Relates to
- https://github.com/Kozea/Radicale/issues/2081
- https://github.com/Kozea/Radicale/pull/2002


pbiering:
@flesueur - I've extended test cases and found that your implementation is also breaking softlinks supported on Windows with same issue found originally on Linux during PROPFIND request, see e.g. this particular workflow run: https://github.com/Kozea/Radicale/actions/runs/24395680534/job/71253826211#step:5:731

Having your change from https://github.com/Kozea/Radicale/pull/2002 reverted by https://github.com/Kozea/Radicale/pull/2087/changes/94746cfb6e0edf9c83d8d3a129b7a5d48241bdfb it is working again and passing the test on Windows https://github.com/Kozea/Radicale/actions/runs/24409725117

Can you please investigate further on, if there is no proper solution then the performance impact would be only on filesystems detected during startup as not collision-free.

flesueur:
Sorry it was much more complex than what I expected ! I guess [this part](https://github.com/Kozea/Radicale/blob/master/radicale/pathutils.py#L294) is still WIP but, anyway, I did not find any other way to solve it.
As soon as we use `realpath()` symlinks are problematic on every OS and surprisingly, there does not seem to be a `realpath()` alternative suited to this need. Anyway, could not find such a workaround.

Even going back to the [original motivation](https://github.com/Kozea/Radicale/commit/2d170bd41fec93aae9a35dbf68aa6d758f10670d) I don't spontaneously find another good alternative :( .

Thanks !

pbiering:
> Sorry it was much more complex than what I expected ! I guess [this part](https://github.com/Kozea/Radicale/blob/master/radicale/pathutils.py#L294) is still WIP

I think there is no other way then predetecting and making it faster by skip on unproblematic file system while keeping the old code for others.

If you agree the block below can be removed again, anyhow forced to be not active at the moment:
https://github.com/Kozea/Radicale/blob/7d07962592697727c8a81da8bdf558ec7536d2a5/radicale/pathutils.py#L295

flesueur:
Yes I think so, it is now useless and can be removed.

Issue #1980, opened 2026-02-03
how to best deal with `libpass` vs. `passlib`


henning-schild:
I tried bringing the latest version to gentoo Linux and fount that `libpass` is not yet packaged there. So without patching Radicale in any way this effort has to wait.

There seems to be a `sed` trick but not sure this is officially supported/tested and what is has to do with bcrypt-5.0 exactly.

https://github.com/Kozea/Radicale/blob/f2f650f9227fef8be2306b4abd7c7142e30b9cef/pyproject.toml#L31

I see @pbiering also active in https://github.com/notypecheck/passlib/issues/11 and https://bugzilla.redhat.com/show_bug.cgi?id=2427085 which seems to indicate that in the redhat world latest Radicale also runs into `libpass` vs `passlib` issues.

Since bcrypt in gentoo is an optional runtime feature enforcing its version would be hard, and introducing some downstream `sed` seems hacky.

Can you please tell me how to best deal with the problem? I would in fact suggest making Radicale officially still support `passlib`, hopefully with bcrypt>=5 and wait for `libpass` to hopefully take over on Pypi.

henning-schild:
https://bugs.gentoo.org/968579

pbiering:
"passlib" is stale, but still packaged on EL and other, would mean as long as "libpass" starts not breaking APIs (which I hope), both can be used...but for "bcrypt" >= 5 "libpass" is a strong requirement.

If "libpass" start breaking APIs then current plan would be defining max version in Radicale config.

henning-schild:
Can we break up the "bcrypt >= 5" needs "libpass" somehow? Maybe you can point me to the code which "breaks" and can tell me why it will/can break.

I can help sorting things out and writing a test matrix while we all wait for "libpass" to become the followup of "passlib" (hopefully).

In gentoo we are waiting for someone to package "libpass", which at the moment waits for it to take over "passlib" on pypi. Maybe we can speed that up and not wait for pypi but will have to see. Having to `sed` patch Radicale is not ideal in the meantime, and forcing bcrypt down to an old version also is not. Hence i am trying to find a way to make Radicale work with all possible combinations so i can ship the most recent version.

But that migration path will likely end up being sort of an "if COMP.version > X; then; else" code ending up in Radicale, together with testing in a matrix of combinations

pbiering:
related code checking whether found _bcrypt_ is supported by found _libpass_/_passlib_ starts here
[...]

PR #2089, opened 2026-04-14
deal with bcrypt>=5 72 chars limit for libpass+passlib

henning-schild:
This introduces truncation on the side of radicale when an overlong password is provided. It also add a testcase which shows that a password with 72+ chars will (and has with bcrypt < 5) result in the same hash a 72 chars. This makes sure bcrypt in fact just truncated before and it is fine to do so.

On top are a few hackish commits to show that users can now use passlib+bcrypt5 again, libpass still works as well and any combination with bcrypt4 should also work.

With this `passlib_libpass_supports_bcrypt` can maybe be retired and passlib+bcrypt5 could become an allowed combination. https://github.com/Kozea/Radicale/issues/1980

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
 ]
 urls = {Homepage = "https://radicale.org/"}
 requires-python = ">=3.9.0"
-# Hint: if bcyrpt < 5.0.0 is used, passlib(libpass) dependency can be downgraded/reverted by: sed -i 's|libpass[^"]*|passlib|' pyproject.toml
+# Hint: passlib(libpass) dependency can be downgraded/reverted by: sed -i 's|libpass[^"]*|passlib|' pyproject.toml
 dependencies = [
     "defusedxml",
     "libpass>=1.9.3",
--- a/radicale/tests/test_auth.py
+++ b/radicale/tests/test_auth.py
@@ -191,6 +191,21 @@ class TestBaseAuthRequests(BaseTest):
         hash = bcrypt.hashpw(("a" * self.BCRYPT_MAX_PWLEN).encode(), bcrypt.gensalt())
         self._test_htpasswd("bcrypt", "tmp:" + hash.decode('utf-8'), m)
 
+    @pytest.mark.skipif(has_bcrypt == 0, reason="No bcrypt module installed")
+    @pytest.mark.skipif(not utils.passlib_libpass_supports_bcrypt()[0], reason="bcrypt module incompatible with passlib(libpass) module")
+    def test_bcrypt_hash_long(self) -> None:
+        import bcrypt
+        salt = bcrypt.gensalt()
+        h0 = bcrypt.hashpw(("a" * self.BCRYPT_MAX_PWLEN).encode(), salt)
+        try:
+            h1 = bcrypt.hashpw(("a" * (self.BCRYPT_MAX_PWLEN + 1)).encode(), salt)
+        # bcrypt >= 5, truncated for us
+        except ValueError:
+            return
+
+        # overlong password produced same hash as maxlen
+        assert h0 == h1
+
     @pytest.mark.skipif(has_argon2 == 0, reason="No argon2 module installed")
     def test_htpasswd_argon2_i(self) -> None:
         self._test_htpasswd("argon2", "tmp:$argon2i$v=19$m=65536,t=3,p=4$NgZg7F1rzRkDoNSaMwag9A$qmsvMKEn5zOXHm8e3O5fKzzcRo0UESwaDr/cETe5YPI")
--- a/radicale/tests/test_auth.py
+++ b/radicale/tests/test_auth.py
@@ -42,6 +42,8 @@ class TestBaseAuthRequests(BaseTest):
 
     """
 
+    BCRYPT_MAX_PWLEN = 72
+
     # test for available bcrypt module
     try:
         import bcrypt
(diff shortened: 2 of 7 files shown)


radicale/tests/test_auth.py
```diff
@@ -144,6 +146,21 @@ def test_htpasswd_bcrypt_2a_autodetect(self) -> None:
     def test_htpasswd_bcrypt_2b(self) -> None:
         self._test_htpasswd("bcrypt", "tmp:$2b$12$7a4z/fdmXlBIfkz0smvzW.1Nds8wpgC/bo2DVOb4OSQKWCDL1A1wu")
 
+    @pytest.mark.skipif(has_bcrypt == 0, reason="No bcrypt module installed")
+    @pytest.mark.skipif(not utils.passlib_libpass_supports_bcrypt()[0], reason="bcrypt module incompatible with passlib(libpass) module")
+    def test_htpasswd_bcrypt_2b_long(self) -> None:
```
henning-schild: should probably move down and not be called "2b", but that is the version which made the char limit clear

radicale/tests/test_auth.py
```diff
@@ -144,6 +146,21 @@ def test_htpasswd_bcrypt_2a_autodetect(self) -> None:
...
+             # full available len
+             ("tmp", "a" * self.BCRYPT_MAX_PWLEN, True),
+             # longer than valid -> same password
+             ("tmp", "a" * (self.BCRYPT_MAX_PWLEN + 42), True),
+             # shorter that before -> some other password
+             ("tmp", "a" * (self.BCRYPT_MAX_PWLEN - 1), False),
+             )
+        hash = bcrypt.hashpw(("a" * self.BCRYPT_MAX_PWLEN).encode(), bcrypt.gensalt())
+        self._test_htpasswd("bcrypt", "tmp:" + hash.decode('utf-8'), m)
+
```
henning-schild: Just to be extra-sure i could add another test not just for `checkpw` but also for `hashpw`

```
try:
  h0 = hashpw(72+)
# bcrypt>=5 telling us to truncate
catch ValueError:
  success
h1 = hashpw(72)

# bcrypt < 5 truncated for us
if h1 == h0:
  success

fail
```

radicale/tests/test_auth.py
```diff
@@ -42,6 +42,8 @@ class TestBaseAuthRequests(BaseTest):
 
     """
 
+    BCRYPT_MAX_PWLEN = 72
```
pbiering: can you not refer to the value in htpasswd.py?

radicale/utils.py
```diff
@@ -112,6 +112,7 @@ def vobject_supports_vcard4() -> bool:
 def passlib_libpass_supports_bcrypt() -> Tuple[bool, str]:
     """Check if passlib(libpass) version supports bcrypt version."""
     info = ""
+    return (True, "accept any version for now")
```
pbiering: is this temporary or can the whole function be removed?

pbiering (changes requested):
please reply

pyproject.toml
```diff
@@ -28,10 +28,9 @@ classifiers = [
 ]
 urls = {Homepage = "https://radicale.org/"}
 requires-python = ">=3.9.0"
-# Hint: if bcyrpt < 5.0.0 is used, passlib(libpass) dependency can be downgraded/reverted by: sed -i 's|libpass[^"]*|passlib|' pyproject.toml
 dependencies = [
     "defusedxml",
-    "libpass>=1.9.3",
+    "passlib",
```
henning-schild: I think for this guy a good choice would be to stay with libpass and include at least one passlib run into actions, or maybe double the full test matrix for both libs

radicale/tests/test_auth.py
henning-schild: added as its own patch, can always be not merged

radicale/utils.py
henning-schild: at the moment still temporary, maybe you review by commit in order and not overall changes, the story and titles should give a clue

henning-schild: The idea is to drop it for good, but before i continue i wanted to get a feeling if the whole story has a chance to be merged at all, before cleaning up all corners

pyproject.toml
pbiering: dedicated tests for passlib and libpass>=1.9.3 would be very good, question is how to implement this proper in the workflow as dependencies need to be changed...potentially a manipulation of pyproject.toml before tox is started?

radicale/tests/test_auth.py
henning-schild: i first wrote the test and later the fix for it ... TDD ;)

But yes it is a magic constant that should be defined only once, will do somehow.

pyproject.toml
henning-schild: Here the idea would be to respect your decision to move to libpass, and just add at least one test that also runs `sed` to switch back and do a compat check with passlib.

I will have to look into github actions for that.

Please let me know how you feel about testing both, staying with "libpass" per default and if one or whole matrix would be best. Whole matrix would double and likely increase test times.

pbiering: > Here the idea would be to respect your decision to move to libpass, and just add at least one test that also runs `sed` to switch back and do a compat check with passlib.

Sounds good.

> I will have to look into github actions for that.
> 
> Please let me know how you feel about testing both, staying with "libpass" per default and if one or whole matrix would be best. Whole matrix would double and likely increase test times.

Let's keep "libpass" for now, no clue what happens if we suddenly revert back...

Regarding test matrix imho it would be enough to duplicate "test-ubuntu-python-newest" and rename it to "test-ubuntu-python-newest-with-passlib", no need to run through the whole matrix. You can chain it dependency wise in parallel to the 32bit test.

henning-schild: > no clue what happens if we suddenly revert back...

Nothing! They are 100% compatible, remember ;) Even if, one thing at a time and libpass is very likely fine for venvs/docker and possibly distros shipping that.

I will add that one job you suggested.

radicale/utils.py
henning-schild: cleaned it up and hopefully all traces left in comments and such

henning-schild:
If that goes green it might be done. Note that i used "Closes:" tags so merging will close issues.

But no rush, and i can remove the tags if needed. And of cause deal with any follow-up change requests of any sort.

pbiering (approved):
thank you very much!

pbiering:
BTW: Fedora 44 finally received an update to 1.9.3 and added "python*dist" provides to new library name.

https://bugzilla.redhat.com/show_bug.cgi?id=2427085

pbiering (approved):
Thank you very much!

henning-schild:
Thank you very much for accepting!

henning-schild:
> BTW: Fedora 44 finally received an update to 1.9.3 and added "python*dist" provides to new library name.
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=2427085

Gentoo was fast to follow:

https://bugs.gentoo.org/969227

Funny how things play out, did a bunch of compat work yesterday and now it is not needed anymore ;). But others might need it.

PR #2090, opened 2026-04-15
setup: drop leftover dependency "packaging"

henning-schild:
Fixes: 6690e9e3ca80 ("radicale/utils: drop bcrypt compat check")


pbiering (approved):
Makes sense, thank you!

PR #2092, opened 2026-04-16
UI: Only clear errors on successful login

maxberger:
The last patch was overly eager to clear errors, we should only clear them on successful login. Added an integ test to make sure this does the right thing.


pbiering (approved):
Thank you!

Issue #2085, opened 2026-04-12
Missing tutorial/Basic Configuration of sharing configuration


srkunze:
Currently cleaning up my old radicale config. While upgrading, I discovered the new sharing feature! 👍 
Right now, I try to make sense of the [sharing] section.

By trial and error, I discovered how to activate it since I could not find anything introductory in the docs or SHARING.md. I think a  tutorial/Basic Configuration/few sentences in SHARING.md could help.

Would you accept a pull request?

AuspeXeu:
Have you managed to share using the web ui? It's a shame such a great feature has no better documentation.

pbiering:
> Have you managed to share using the web ui? It's a shame such a great feature has no better documentation.

Feel free to contribute to Wiki page: https://github.com/Kozea/Radicale/wiki/Sharing-Collections

Me (for the backend) and @maxberger (WebUI) had first to concentrate on implementation and bugfixes.

How it works via API is already documented: https://github.com/Kozea/Radicale/blob/master/SHARING.md#sharing-access-via-maps

pbiering:
> Would you accept a pull request?

Depends on the content, partially it can be better to extend the existing wiki page https://github.com/Kozea/Radicale/wiki/Sharing-Collections or create even a new one.

srkunze:
> Have you managed to share using the web ui? It's a shame such a great feature has no better documentation.

Yes, I did!

You know like we seasoned coders always do such things: mediating over the config file, finding and reading various pieces of documentation, AND, after a large cup of coffee, I finally just randomly turned config variables on and off and refreshed the Web UI frequently. 😂

srkunze:
@pbiering I see that there multiple docs places for this feature. I found the following:

https://radicale.org/v3.html#sharing
https://github.com/Kozea/Radicale/blob/master/SHARING.md
https://github.com/Kozea/Radicale/wiki/Sharing-Collections

I initially found the first link as this is what Google finds and is referred to frequently all over the Internet.

> Depends on the content, partially it can be better to extend the existing wiki page [...] or create even a new one.

From a consumer perspective, I would do the following

- add a subsection for sharing to https://radicale.org/v3.html#basic-configuration
- move/merge SHARING.md to Wiki
- add a new section for a short explanation of the Web UI on https://radicale.org/v3.html

Whatever you approve, I can do.

pbiering:
Github Wiki content is not part of package, so potentially also not the best destination for API and WebUI documentation.

[...]

PR #2103, opened 2026-04-23
First stage of #2085 - split DOCUMENTATION.md

srkunze:
- major changes to run.py
  - main contribution are the following four functions
    - ensure_clean_workspace cleans the repository of all previous artifacts
    - collect_docs_sources restores the docs/ folder from release branches
    - build_docs_html merges all MD files into one HTML file, then extracts smaller ones congruent with the MD file structure
    - publish_html comprise the staging, committing and pushing to gh-pages which then lead to an update to the webpage hosted by Github
  - performance optimizations
    - added class MDFile for cleaner data-recycling like target html file path or headings
    - replace multiple calls to git by a single `git restore`, also not to mess up the repository working directory and staging area
    - substitute the subprocess call to prostprocessor.py by a regular function call
    - substitute the call to filter.py (by Pandoc) by a regular function call
  - extend postprocessor to augment all href by html filename, so switching between files produces no broken links
  - optimize imports
  - optimize constants
  - reformat
  - make sure sys.executable is always used
  - validate if duplicate level2 headings exists
- improve README.md
- fix asset links in template.html
- remove template-redirect.html, redirect.json; no need for special treatment - everything needs to be in the respective git branch
- fix href selector in navigation.js to match by file and anchor
- fix static links in 1.1.html, 2.1.html, 3.3.html
- remove single-file documentation artifacts v1.html, v2.html, v3.html, master.html; they are substituted by html/v1/, html/v2/, html/v3/ and html/master/


srkunze:
@pbiering Alright. This is the main pull-request.

We need to pull over those other 4 branches as well. Do you want me to open pull-requests as well for these?

pbiering:
@srkunze - thank you for all the work, but can you shrink the PR somehow?

v1 and v2 doc I would leave untouched, there is no change planned anymore, they are there for historic reasons only.

v3 is after each release a rebase of the master branch. HTML pages are generated on each related documentation commit, see 

https://github.com/Kozea/Radicale/blob/master/.github/workflows/generate-documentation.yml

So imho it would be enough filing 2 PRs

For branch "gh-pages":

- updated generator able to handle old single doc file (v1/v2 - in case of any changes) or splitted doc (master -> v3)
- fixes for legacy files (potentially an extra commit)

For branch "master"

- splitted and relocated md doc files
- extended github workflow to run on push, e.g. 

```
on:
  push:
    paths:
     - DOCUMENTATION.md
     - docs/*.md
```

After next push to "master" (or manual workflow trigger), the generator should start and create for "master" multi-html pages and all can be tested while keeping v1/v2/v3 doc still on as-is.

On next v3 release with rebase to "master" the generator will also be triggered and create splitted v3 doc.

Does this work-out?

srkunze:
@pbiering Good catch about the missing diff in `.github/workflows/generate-documentation.yml`.

Will test the workflow and get back to you soon.

srkunze:
@pbiering Alright. I tested the github workflows: https://github.com/srkunze/Radicale/actions/workflows/generate-documentation.yml

Seem to work fine after some debugging.

---

I also made up my mind about not changing v1 and v2. There are two big issues:

1) The `assets/` were changed in order to work with folder-based HTML documentation. These simply don't work out of the box with the previous version of HTML files any more. As docs span multiple version, this probably worsens maintenance.
2) Future developments -- if they touch on v1 or v2 -- will be able to change docs easier because it's managed equally across branches.

The more I think about it, the more places in `run.py` would need to be touched as well and more special cases would be introduced for v1 and v2 only. I'd rather not increase the file size (and maintain it indefinitely) for something that can be fixed by two additional pull requests.

So, to answer the question

> "can you shrink the PR somehow?"

So unfortunately, no. The work for deprecated versions won't be worth it. Let's clean it up in a one-time action.

srkunze:
Leaving out v3 PR would work, because we would have failing Github workflows until everything is merged in any case (i.e. until you release a new version by rebasing v3 onto master in this case). If we want to avoid that, merging the v3 PR would resolve that instantly.

pbiering:
will the PR get additional adjustments or simply merge and check what's happen?

srkunze:
@pbiering From my understanding that's it. Merge as is, no conflicts, no adjustments and done.

Next step is then to add tutorial for sharing.

If you agree, I would open the respective pull requests (and close this one here as it's not the right branch). According to `git log` all commits can be merged without conflicts (as of Yesterday). I think I won't have much time on the weekend. But I can plan a time window on Monday.

pbiering:
@srkunze - I feel the PR is still too complex.

There is beside the README.md not any documentation md file in the "gh-pages" branch.

I would expect 2 PRs for the different branches

- "gh-pages": add extended generator and assets
- "master": splitted md files and updated workflow file related to generator trigger incl. extension of `on` section (see above)

Next steps

- trigger the generator by workflow, which creates then html files in html/master directory
- replace master.html content with redirect to html/master/index.html 

Note: keep the v1.html v2.html v3.html as is for now

On next "v3" release by rebasing to tag of "master", the generator is also triggered and create html files in html/v3 directory

Execute manual change on gh-pages branch and align v3.html to point to html/v3/index.html 

I would also prefer that the new generator will have a different name, e.g. "run-splitted.py" and refer this in the "master" github workflow only:

```
      - name: Run generator
        run: documentation-generator/run-splitted.py
```

Goal should be having both available in parallel to avoid the need of touching v1 and v2 doc and able to revert easily also for "master" in any unexpected case.

One can think of that the new generator will create master.html or v3.html automatically with the redirect content, while the old generator creates the one-pager file.

Once this is working, you can create a PR to the "master" branch in doc directory with a new section inserted related to sharing, which leads to rename of the following doc files, but should not be a big problem, seldom happened....otherwise one would need a sequence info file to define the order of the md files.

As it can be that there will be pictures (e.g. screen shots) inserted into the doc, please check whether the new generator is already prepared for that taking referenced pictures from doc/images/ subdirectory.

Thank you!

pbiering:
@srkunze - have you given up or can I expect simplifications still?

Goal should be as of now:

- keep "v1" and "v2" untouched and having only an additional generator active for master" (and once confirmed working for "v3" only)
- PR to "gh-pages" additional generator - should not contain any generated HTML files as they will be generated by the action itself
- PR to "master": splitted md files in "master" tree (1st shot: in parallell to the main doc, can be autogenerated from this) + workflow file update for activating additional generator only on "master" as of now (to avoid that it's disturbing any new "v3" release

PR #2108, opened 2026-04-25
UI: Add support for usernames with @ symbol

### LINKED ISSUE #2101: The sharing breaks with usernames that have a '@'
The sharing API uses a PATH_PATTERN regex that does not include @, which breaks sharing for any user whose username contains an @ (common with email-based authentication like Dovecot/LDAP). The web UI returns a 400 Bad Request with "unsupported PathMapped" when trying to list or create shares.

I'm using 3.7.1

maxberger:
Fixes #2101

### DIFF
--- a/integ_tests/common.py
+++ b/integ_tests/common.py
@@ -47,12 +47,22 @@ class Config:
     auth_type: AuthType
     sharing_type: SharingType
     extra_config: str = ""
+    admin_username: str = "admin"
+    user_username: str = "max"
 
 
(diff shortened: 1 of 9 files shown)


maxberger:
The same test succeeded both on my local machine and on my fork: https://github.com/maxberger/Radicale/actions/runs/24938780541

I need to see what is happening here.

pbiering (approved):
thank you very much!

Issue #2097, opened 2026-04-21
Shared calendar appears in Thunderbird but not DAVx5


fpusan:
Thanks for the great tool and the continued support!
I am testing the new sharing functionality and found what could be a bug, or an error in my part...

### Background

I created a calendar with User1, and shared it with User2. I could not make it through the web UI since the real usernames contained dashes, which somehow resulted in invalid requests. So I created the `data/collections/collection-db/sharing.csv` file manually:

```
ShareType;PathOrToken;PathMapped;Conversion;Owner;User;Permissions;EnabledByOwner;EnabledByUser;HiddenByOwner;HiddenByUser;TimestampCreated;TimestampUpdated;Properties;Actions
map;/User2/9bab3765-c28f-1d1b-4dec-f5e93f2e7af3/;/User1/ae4f0768-b5aa-c678-fb2c-ee0441c8ab7e/;none;User1;User2;rwP;True;True;False;False;1776774355;1776774355;{};
```

### Behaviour

On Thunderbird, I can add the calendar with both users, as well as viewing, creating and editing events.

With DAVx5 I can add the calendar with both users, but I can only see events with the original owner (this using Android + Google Calendar).
Strangely enough, I can create new events with the user receiving the share. The event stays there until next time DAVx5 syncs, then it disappears from that user, but actually gets added to the calendar.

### Config

In case it helps, my config is:

```
[auth]
type = http_x_remote_user
[web]
type = internal
[storage]
predefined_collections = {
    "personal-addressbook": {
       "D:displayname": "Personal Address Book",
       "tag": "VADDRESSBOOK"
    },
    "personal-calendar": {
       "C:supported-calendar-component-set": "VEVENT,VJOURNAL,VTODO",
       "D:displayname": "Personal Calendar",
       "tag": "VCALENDAR"
    }
  }
[rights]
type = owner_only
[sharing]
type = csv
collection_by_map = true
permit_create_map = true
permit_properties_overlay = true
default_permissions_create_map = rw
```

pbiering:
Hmm, will try to replicate.

Regarding username with `-` you can add `\\-` into regular expression then it should work (will be fixed with 3.7.2, also `_` was missed)

https://github.com/Kozea/Radicale/blob/7d07962592697727c8a81da8bdf558ec7536d2a5/radicale/sharing/__init__.py#L126

pbiering:
I tried a partial replay

- having a group calendar in a group directory
- creating a share to a user

```
[...]

PR #2109, opened 2026-04-25
UI: Trim owner length for incoming shares

maxberger:
This should ensure icons are visible in cases with overly long owner names such as in [#2097](https://github.com/Kozea/Radicale/issues/2097)


pbiering (approved):
thank you very much!

PR #2111, opened 2026-04-26
UI: Fix handling of servers with Prefix

maxberger:
As reported in https://github.com/Kozea/Radicale/issues/2097#issuecomment-4316286515


pbiering (approved):
Thank you!

PR #2112, opened 2026-04-26
UI: Fix: Load username from principal component

maxberger:
Another one to make life easier for people who had the issue mentioned in https://github.com/Kozea/Radicale/issues/2097#issuecomment-4316286515

This fixes an issue where PROPFIND / fails due to strict security settings, but the actual user is still given via x_http_user.


pbiering (approved):
Thank you

PR #2116, opened 2026-04-29
expose server version via PROPFIND and display in UI

maxberger:
This only works for authenticated users to prevent abuse; using a special RADICALE:VERSION tag to ensure we stay compatible to standard servers. 

-> Please decide if we actually want this feature, and feel free to reject this pull request if we don't


pbiering (approved):
Makes sense

pbiering:
Thank you, Wiki extended: https://github.com/Kozea/Radicale/wiki/Radicale-Custom-DAV-protocol-extensions

PR #2117, opened 2026-04-29
Fix: Avoid race condition in integ test

maxberger:


pbiering (approved):
Ah, potentially the reason why fails sometime and re-run was successful

PR #2118, opened 2026-04-30
UI: Re-use sharing cache for individual shares

maxberger:
This should reduce network usage in some situations.


pbiering (approved):
Thank you!

PR #2120, opened 2026-05-02
tiny documentation correction

madblobfish:
When checking out the documentation for the new features I saw that the default value does not exist for `validate_user_value`:
https://radicale.org/master.html#validate_user_value
same with `validate_path_type`.

This fixes the docs, thanks for Radicale and have a great day!


pbiering (approved):
Thank you for catching a last-minute change not reflected in the doc!

PR #2129, opened 2026-05-16
Fix bday conversion hrefs for vCard filenames ending in v/c/f

joinemm:
The bday conversion maps .vcf hrefs to .ics hrefs using `rstrip(".vcf")`.

In Python, `rstrip(".vcf")` removes any trailing characters from the set `., v, c, f` instead of just removing the literal `.vcf` suffix.

For example:
```
"21757fa1-03d6-4cd6-a20a-7b581c7ad9cc.vcf".rstrip(".vcf") + ".ics"
```
produces:

```
21757fa1-03d6-4cd6-a20a-7b581c7ad9.ics
```

instead of:
```
21757fa1-03d6-4cd6-a20a-7b581c7ad9cc.ics
```

This breaks generated birthday calendar items when the source vCard basename ends in `v`, `c`, `f`, or `.`.

The fix is to use `removesuffix()` instead.

### DIFF
--- a/radicale/app/propfind.py
+++ b/radicale/app/propfind.py
@@ -159,7 +159,7 @@ def xml_propfind_response(
         if uri.startswith(share['PathMapped']):
             uri = str(share['PathOrToken']) + uri.removeprefix(share['PathMapped'])
         if share_bday_automap and uri.endswith(".vcf"):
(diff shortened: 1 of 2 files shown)


pbiering:
Thank you for catching this!

PR #2134, opened 2026-05-21
UI: Avoid redundant URL completion in CollectionsScene

bb2:
**Problem Description:**
Redundant URL completion in `CollectionsScene.js`. The file passes a `collection.href` that already contains the full path prefix to `UrlTextHandler.setHref`. Because `setHref` internally calls `completeHref` again, the path is double-completed (over-completed), resulting in a malformed URL.

**Solution:**
Instead of using `completeHref(collection.href)`, use `window.location.origin + collection.href` to manually construct the full URL, preventing the secondary completion bug.


pbiering (approved):
Ok

pbiering:
@bb2 JS test failed, please investigate.

pbiering (approved):
Thank you

PR #2135, opened 2026-05-21
Web UI dark theme with automatic browser/OS color scheme detection

royborgen:
This merge request adds a dark theme for the web UI with automatic theme switching based on the user’s browser and operating system preferences.

The `main.css` has been reworked and I've implemented CSS variables to support both light and dark modes. All color styling is now handled through variables.

Theme selection is handled using the `prefers-color-scheme` media feature that is baked into all modern browsers.

Behavior:
- Users who prefer light mode will get the default light styling. 
- Users who prefer dark mode will get the new dark styling. 
- If the browser is configured to follow system settings automatically, the operating system preference will determine which theme is loaded.

This implementation works with modern browsers that support prefers-color-scheme.

Tested Browsers:
- Chromium 148.0.7778.178
- Firefox 151.0
- IronFox 151.0 
- Vanadium 148.0.7778.178


pbiering:
@royborgen is it not possible to simplify this as this ends up in need to maintain 2x CSS files in parallel keeping in sync for non dark/light mode changes.

royborgen:
hmmm, yes should be doable using variables in CSS. Will look more into it.

royborgen:
@pbiering: I’ve refactored the styling back to a single CSS and implemented CSS variables to support both light and dark modes. All color styling is now handled through variables.

pbiering:
@royborgen is new the empty line 769 intended? If not, please remove.

radicale/web/internal_data/css/main.css
```diff
@@ -641,4 +765,5 @@ button.inline {
 
 .hidden {
     display: none !important;
-}
\ No newline at end of file
+}
+
```
pbiering, line 768: Please remove empty line

radicale/web/internal_data/index.html
```diff
@@ -407,4 +407,4 @@ <h1 data-name="headertitle">Delete Collection</h1>
   </main>
 </body>
 
-</html>
\ No newline at end of file
+</html>
```
pbiering: Please keep the original file and leave it untouched.

royborgen:
> @royborgen is new the empty line 769 intended? If not, please remove.

Cleaned up :)

radicale/web/internal_data/css/main.css
```diff
@@ -641,4 +764,4 @@ button.inline {
 
 .hidden {
     display: none !important;
-}
```
maxberger, line 644: please keep new line at the end of the file

radicale/web/internal_data/index.html
```diff
@@ -407,4 +407,4 @@ <h1 data-name="headertitle">Delete Collection</h1>
   </main>
 </body>
 
-</html>
```
maxberger, line 410: ple keep newline. Since there is no change, probably best to revert this file.

radicale/web/internal_data/css/main.css
```diff
@@ -1,6 +1,130 @@
+/* Default (light mode) */
```
maxberger, line 1: Thanks! Unifying all colors in one place is great! I guess one of the next follow-ups (not for now) would then be to see if we can remove some of the colors, because this list is too long.

maxberger (changes requested):
Please fix the newlines, otherwise I really like this!

radicale/web/internal_data/css/main.css
royborgen: I agree that the list is long. When doing this I realized that there were a lot of variables with duplicated values. I kept them separated intentionally so they can be modified individually without affecting other parts of the UI. I felt that this gives more flexibility. It also felt safer when implementing dark mode. 

I tried to keep a sensible naming scheme based on where the color is actually used, rather than naming them based on color. This should however be easy to change if we do decide to do so later.

PR #2138, opened 2026-05-23
UI: Decode URL in text field for better readability

bb2:
**Optimize visual display:**
Decodes the URL displayed in the input field using `decodeURIComponent`. This makes the displayed link more user-friendly and intuitive, avoiding a screen full of percent signs in internationalized scenarios where collection folder names contain non-English characters or special symbols.

**Ensure copy safety:**
Preserves the original, encoded standard URL internally as `_completedUrl`. Clicking the "Copy" button or manually selecting the entire text box to copy will still place the original, encoded standard link into the clipboard.


pbiering (approved):
Thank you

PR #2139, opened 2026-05-23
UI: Decode URL in text field for better readability

bb2:
This PR is a follow-up to #2138. It extends the URL decoding logic to include the delete and edit collection dialogs, as well as homepage collection titles, ensuring a consistent user experience.


pbiering (approved):
Thank you

Issue #2137, opened 2026-05-23
Issue with copy to clipboard button in the Web interface


yuukiAme:
Things I have done before creating this issue
- [x] Checked the repo issue page for the similar issue - not found
- [x] Checked if tomsquest/docker-radicale repo has a similar issue

Information about the environment

- using image: tomsquest/docker-radicale
- Deployed in Portainer
- version 3.7.3

Proof:

<img width="250" height="85" alt="Image" src="https://github.com/user-attachments/assets/a1ae013a-6b16-4789-a929-8a593b44b9d3" />

Issue description:
I **cannot** click on copy to clipboard symbol for the URL. I think the button itself is **unlink**. 

Screenshot here

<img width="305" height="387" alt="Image" src="https://github.com/user-attachments/assets/519c52a0-c254-4053-b9fd-be22516b8594" />

Things I have tried
- using two different browsers (Firefox vs Edge browser)
- disabled all the extensions
- checked if my clipboard or copy function is disabled
- checked if my left click on my mouse is broken

Ask / primary concern:
- Is it a bug?

pbiering:
@maxberger can you please investigate?

maxberger:
Hi,

unfortunately I need a bit more information to investigate this one, since it works for me. When the button is clicked, the icon should shrink a little, and also turn green for a short time. Then the URL should be copied to the clipboard.

* Do you see the animation at all?
* I guess the copy to clipboard is not doing anything? Is that the error?
* Can you check if there are any error messages in the console? Since you have 2 browsers to test, ideally check both and if there are error messages there paste them.
* You say you are running this locally -> Are you using http or https? [Background: There are some clipboard issues with http]

Thanks!

Max

yuukiAme:
> Hi,
> 
> unfortunately I need a bit more information to investigate this one, since it works for me. When the button is clicked, the icon should shrink a little, and also turn green for a short time. Then the URL should be copied to the clipboard.
> 
> * Do you see the animation at all?
> * I guess the copy to clipboard is not doing anything? Is that the error?
> * Can you check if there are any error messages in the console? Since you have 2 browsers to test, ideally check both and if there are error messages there paste them.
> * You say you are running this locally -> Are you using http or https? [Background: There are some clipboard issues with http]
> 
> Thanks!
> 
> Max

Hi, @maxberger ,

Thanks for your attention to this issue.
[...]

PR #2141, opened 2026-05-23
Hide copy button in insecure environments

maxberger:
writing to clipboard via javascript only works in secure environments. Likely cause for Issue #2137


pbiering (commented):
Ok

Issue #2143, opened 2026-05-26
LoginScene auto-login probe: should credentials: 'omit' be configurable for SSO setups?


grasbauer:
LoginScene.show() probes for existing authentication with credentials: 'omit' to avoid the browser's native Basic Auth dialog on 401.

This breaks SSO setups using type = http_x_remote_user with a cookie-based reverse proxy (oauth2-proxy in my case): the session cookie is never sent, the proxy can't inject X-Remote-User, and Radicale returns 401 — so the Login screen appears even though the user is already authenticated.

Changing to credentials: 'include' fixes this. I think modern browsers' SameSite=Lax default already prevents the CSRF concern, and with http_x_remote_user Radicale never sends WWW-Authenticate: Basic so no dialog appears.

Would it make sense to change this to 'include', or add a server-side hint (e.g. a response header) that the web UI could use to decide?

pbiering:
@maxberger - can you please investigate

maxberger:
Thank you very much for this suggestion. In my personal case I use the same setup, and I have not been able to get this to work properly. I will look at credentials:include and see if this works.

The current behavior is that it will accept any username and password, and then overwrite that with the username we get from the sso. So while it should work, it is clunky. However this has been the behavior before I started the latest changes, so I let that one go for the moment.

As you said what I don't want is the browser login to pop up for "the normal cases", so we may need to make this a user defined setting, as you suggested. I'll take a look later today.

grasbauer:
https://github.com/grasbauer/radicale_oauth

this is my playground - possibly I missed something

maxberger:
I proposed the following patch for configuring the setting:
  https://github.com/Kozea/Radicale/pull/2145
Please let me know if you think this would help!

grasbauer:
👍 works like a charm - thank you.

PR #2145, opened 2026-05-27
web ui: New setting for browser-based login

maxberger:
This is useful in cases where OIDC login is active which would then redirect the user to a central
login page.

See https://github.com/Kozea/Radicale/issues/2143


pbiering (approved):
Thank you!

PR #2149, opened 2026-06-01
contrib/ferron: add Ferron example configuration

lvgrr:
Add a minimal configuration for making Radicale work with [Ferron](https://ferron.sh/) web server.

The version of Ferron is 2. Other versions use different file formats, but with version 1.x I was not able to make Radicale work, while version 3 is still in beta.


pbiering:
Thank you!

PR #2156, opened 2026-06-09
Sharing block update conversion

pbiering:
block update of conversion (not supported), align some error messages


pbiering:
@maxberger - please investigate "integ_tests" problem, have not found on first look the root case
https://github.com/Kozea/Radicale/actions/runs/27231663684/job/80413541479 - merged regardless, so check "master" please.

PR #2160, opened 2026-06-11
Remove obsolete integ_test

maxberger:
This test was checking for conversions between sharing configs, which is no longer possible


pbiering (approved):
Thank you!

PR #2162, opened 2026-06-13
UI: disable conversion mode buttons on edit

maxberger:
The conversion type cannot be changed once a share is created, so disable the edit buttons.


pbiering (approved):
Thank you!

PR #2171, opened 2026-06-24
Append domain to login before trying IMAP connection

mbartel:
Append email domain to login before trying connect to the IMAP server.

Use this instead of strip_domain, which removes the domain before trying the IMAP connection, to avoid the domain in the folder names of the collection.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1389,6 +1389,14 @@ One of:
 
 Default: `tls`
 
+##### imap_append_domain
+
+_(>= 3.7.6)_
+
+Append domain to login before trying IMAP connection like @domain.tld
+
+Default: (unset)
+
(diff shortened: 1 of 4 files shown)


pbiering:
@mbartel - thank you for the extension, but would it not better to let default "empty" instead of using "none"?

mbartel:
> @mbartel - thank you for the extension, but would it not better to let default "empty" instead of using "none"?

Yes, I agree. I'm not familiar with python, so I did the "none" version. I will have some time this weekend to try the "empty" version. Hopefully I'll be able to commit and push it until monday.

pbiering:
@mbartel - are you able to fix it or should I overtake?

pbiering:
superseeded by https://github.com/Kozea/Radicale/pull/2176

PR #2174, opened 2026-07-02
add documentation for `groups` in rights file

tendstofortytwo:
This config key is undocumented, seemingly by accident? We ran into this while setting up Radicale for https://queercomputerclub.ca - had to discover the key in source code so we could use it


pbiering:
thank you for catching this!

pbiering (approved):
ok

PR #2178, opened 2026-07-11
Fix: time-range filter matches VEVENT with whole-day DURATION

TowyTowy:
## Problem

A `calendar-query` `REPORT` with a `time-range` filter does not return a `VEVENT` that uses a whole-day `DURATION` (e.g. `DURATION:P1D`, `P2D`) when the queried range falls inside the event but after `DTSTART`. The same event expressed with `DTEND` is returned correctly.

## Cause

In `radicale/item/filter.py`, the VEVENT time-range logic (rfc4791-9.9) selected the "non-zero duration" branch (line 2) using `timedelta.seconds` instead of `timedelta.total_seconds()`. For a duration that is an exact multiple of 24h, `timedelta.seconds` is `0` (the value is in `.days`), so the event was mis-classified as zero-length (line 3) and only matched a one-second window at its start.

## Fix

Use `total_seconds()`. One-line behavioural change plus a regression test.

## Tests

Adds `event11.ics` (`DURATION:P2D`) and `test_time_range_filter_events_whole_day_duration`, asserting an inside-range match and an outside-range non-match. Fails before the fix, passes after. Full suite: 381 passed, 14 skipped. flake8/isort/mypy (CI-pinned versions) clean.

No existing issue tracked this; found while reviewing DURATION handling in the time-range filter. I bumped the CHANGELOG/`pyproject` to a new `3.7.7.dev` section per the convention documented in `pyproject.toml` — happy to drop that if you'd rather own the version.

---
*Disclosure: this change was prepared with AI assistance (Claude). The bug was verified against the code and the RFC, and the fix and regression test were run locally (fail-before / pass-after) before submission.*


pbiering (approved):
Good catch, thank you!

pbiering:
@maxberger - the "JS Type Check" test starts failing: https://github.com/Kozea/Radicale/actions/runs/29155326925/job/86618214847?pr=2178 (but this is unrelated to this PR)

maxberger:
> @maxberger - the "JS Type Check" test starts failing: https://github.com/Kozea/Radicale/actions/runs/29155326925/job/86618214847?pr=2178 (but this is unrelated to this PR)

Thanks, I will take a look. My hypothesis: we are using the typescript compiler for these checks (we don't use type script, but this is very good at checking java script). TS 7 has been released 2 days ago, and probably has additional checks which are now failing. It is suspicious that it is just 2 types or errors. A quickfix would be to just ignore these two types in the as part of the jsconfig, but that would invalidate the point of the checker.

maxberger:
> > @maxberger - the "JS Type Check" test starts failing: https://github.com/Kozea/Radicale/actions/runs/29155326925/job/86618214847?pr=2178 (but this is unrelated to this PR)
> 
> Thanks, I will take a look. My hypothesis: we are using the typescript compiler for these checks (we don't use type script, but this is very good at checking java script). TS 7 has been released 2 days ago, and probably has additional checks which are now failing. It is suspicious that it is just 2 types or errors. A quickfix would be to just ignore these two types in the as part of the jsconfig, but that would invalidate the point of the checker.

https://github.com/Kozea/Radicale/pull/2179

PR #2179, opened 2026-07-12
Fix new JS verification errors and force specific TSC version

maxberger:
The new TSC compiler does stricter checking on the JS documentation strings. All documentation strings have been updated to pass with the current version (7.0.2)

In addition, the version of the TSC compiler used during the github action will be fixed to 7.0.2, so that we don't get these type of sudden errors again in the future. Unfortunately this means we need to periodically update this manually.


pbiering (approved):
Thank you!

PR #2180, opened 2026-07-12
Fix: expand REPORT leaves recurrence properties (RDATE) on expanded instances

TowyTowy:
### Problem

A `calendar-query`/`calendar-multiget` REPORT with `<C:expand>` returns individual recurrence instances. Per RFC 4791 §9.6.5, each returned component *"MUST NOT use recurrence properties (i.e., EXDATE, EXRULE, RDATE, and RRULE)"*.

`_strip_single_event()` removes these via a single `try/except AttributeError` around four sequential `delattr()` calls (`rrule`, `exdate`, `exrule`, `rdate`). When an event lacks one of them — commonly `EXDATE` — the `AttributeError` aborts the block, so the remaining properties, in practice **`RDATE`**, stay on every expanded VEVENT.

### Reproduction

Store a recurring event with `RRULE` + `RDATE` and no `EXDATE`, then request an expand REPORT over the recurrence range: each expanded VEVENT (with its own `RECURRENCE-ID`) still contains the original `RDATE`.

### Fix

Remove each recurrence property independently so a missing one no longer prevents removal of the others.

### Tests

New fixture `event_rrule_rdate.ics` and `test_report_with_expand_property_strips_recurrence_properties`, asserting the non-expanded response still carries RRULE/RDATE while the expanded response carries neither (and keeps RECURRENCE-ID). Fails before the change, passes after; full `test_expand.py` green (25 passed); flake8/mypy/isort clean. CHANGELOG updated.

_Disclosure: found by code inspection (no filed issue) and prepared with AI assistance (Claude); reproduced, authored and reviewed by @TowyTowy._


radicale/tests/test_expand.py
```diff
@@ -750,3 +750,34 @@ def test_report_with_expand_property_all_day_event_overridden(self) -> None:
...
+        recurrence-defining property (RRULE, RDATE, EXRULE, EXDATE).
+        """
+        uid = "event_rrule_rdate"
+        start = "20060102T000000Z"
+        end = "20060105T000000Z"
+
+        # Baseline (no expand): the stored event still holds RRULE and RDATE.
+        _, responses = self.report(
+            "/calendar.ics/", self._req_without_expand(uid, start, end))
+        status, element = responses[
```
pbiering: mypy test reports: 
radicale/tests/test_expand.py:769: error: Value of type "int | dict[str, tuple[int, Element]] | Any" is not indexable  [index]

radicale/tests/test_expand.py
```diff
@@ -750,3 +750,34 @@ def test_report_with_expand_property_all_day_event_overridden(self) -> None:
...
+            f"/calendar.ics/{uid}.ics"]["C:calendar-data"]
+        assert status == 200 and element.text
+        assert "RRULE" in element.text
+        assert "RDATE" in element.text
+
+        # Expanded: individual instances must not carry recurrence properties.
+        _, responses = self.report(
+            "/calendar.ics/", self._req_with_expand(uid, start, end))
+        status, element = responses[
```
pbiering: mypy test reports: radicale/tests/test_expand.py:778: error: Value of type "int | dict[str, tuple[int, Element]] | Any" is not indexable  [index]

pbiering:
Thank you for reporting, but please fix the test extension, I assume a failsafe check above the mentioned lines is required to make "mypy" happy.

TowyTowy:
Good catch, thanks. Fixed — I assign the response to a variable and `assert isinstance(response, dict)` before indexing it, matching the pattern already used elsewhere in `test_expand.py` (e.g. the `test_report_with_expand_*` tests). `mypy` now passes on the file, and the test itself is unchanged and still green.

pbiering (approved):
Thank you

PR #2181, opened 2026-07-15
Fix: text-match filter crashes on structured property (vCard N/ADR)

TowyTowy:
A CardDAV `addressbook-query` REPORT that applies a `text-match` prop-filter to a structured property such as `N` or `ADR` returned HTTP 500. vobject parses those into `Name`/`Address` objects rather than plain strings, so `text_match()` called `.lower()` on a non-string and raised `AttributeError` (surfaced as a 500 via `xml_report`'s exception wrapping).

Fix: coerce non-string property values to their text representation (`str(value)`, e.g. `"Dr John Q Doe Jr"`) before matching. String and list-valued (CATEGORIES/ORG) properties are unaffected. Added a regression test and a CHANGELOG entry.

Verified: the new test fails before the change (500 / `'Name' object has no attribute 'lower'`) and passes after; broader `addressbook`/`filter`/`text_match` tests (28) pass; flake8/isort/mypy clean.

_Prepared with AI assistance (Claude) and reviewed/verified by me._


pbiering (approved):
Thank you very much!

PR #2182, opened 2026-07-15
Fix: bday conversion clobbers NICKNAME when FN is empty

TowyTowy:
The share-by-map BDAY -> ICS conversion builds a placeholder mapping for the SUMMARY/DESCRIPTION/alarm templates. Each block sets the fallback marker for its own placeholder (`{nickname}` -> `!nickname!`, `{n:f}` -> `!n:f!`, ...), but the `{fn}` block wrote the empty-FN fallback to `{nickname}` instead of `{fn}`.

For a VCARD whose `FN` is present but empty, a genuine `NICKNAME` was therefore overwritten with `!fn!` in the generated event, and `{fn}` never received its `!fn!` marker, so a `[{fn}|...]` fallback in a template failed to resolve. This assigns the fallback to `{fn}`, matching the surrounding blocks.

Added a regression test (`test_sharing_bday_conversion_empty_fn`) that fails before and passes after, plus a CHANGELOG entry.

Prepared with AI assistance (Claude Code) and reviewed/verified by me.


pbiering (approved):
good catch, thank you!

pbiering:
@TowyTowy - are you ongoing looking for other bugs or when do you expect having finished your first round of bug-hunting? I'm asking because of potential shift of release date of 3.7.8.

PR #2183, opened 2026-07-16
Fix: free-busy REPORT always fails with HTTP 400 when max_freebusy_occurrence is 0

TowyTowy:
**Problem**
With `[reporting] max_freebusy_occurrence = 0` (limit disabled), every `free-busy-query` REPORT on a calendar containing items fails with HTTP 400 and logs `FREEBUSY occurrences limit of 0 hit`.

**Cause**
`free_busy_report()` already treats `max_occurrence == 0` as "no limit" when fetching occurrences (`n_occurrences = 0` lets `time_range_fill()` return everything, see 408a03a3), but the subsequent check `if len(occurrences) >= max_occurrence:` is trivially true for `max_occurrence == 0`, so it always raises.

**Fix**
Skip the limit check when the limit is disabled (`max_occurrence > 0 and ...`), consistent with how `xml_report()` handles the same setting (`if max_occurrence and n_vevents > max_occurrence`). Behavior for positive limits is unchanged (the existing `max_freebusy_occurrence = 1` test still expects 400).

**Testing**
Extended `test_report_free_busy` with the `max_freebusy_occurrence = 0` case (expects 200 and the same 3 VFREEBUSY components as the default-config case). Fails before the fix, passes after; full `test_base.py` suite green (152 passed); flake8/isort/mypy clean on touched files. CHANGELOG entry added under 3.7.7.dev.

Disclosure: this fix was prepared with the assistance of an AI tool (Claude Code) and reviewed/verified by me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 3.7.7.dev
+* Fix: free-busy REPORT always failed with HTTP 400 ("FREEBUSY occurrences limit of 0 hit") when [reporting] max_freebusy_occurrence is set to 0 (limit disabled), because the limit check did not honor the disabled limit
 * Fix: time-range filter treated a VEVENT with a whole-day DURATION (e.g. P1D, P2D) as zero-length (timedelta.seconds instead of total_seconds), so such events were missing from calendar-query REPORT results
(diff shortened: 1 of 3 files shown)


pbiering (approved):
thank you very much!

PR #2184, opened 2026-07-18
Fix: serve_resource/serve_folder ignore their mimetypes and fallback_mimetype parameters

TowyTowy:
`httputils.serve_resource()` and `serve_folder()` accept `mimetypes` and `fallback_mimetype` parameters and pass them into `_serve_traversable()`, but the Content-Type lookup there uses the module-level `MIMETYPES` / `FALLBACK_MIMETYPE` constants, so the parameters are silently ignored.

The parametrized helper was introduced in 33fcda7c ("Extract httputils.serve_folder") for use by custom web plugins; the sibling parameters `path_prefix` and `index_file` are honored, but the extracted body kept the constants from the pre-extraction code. A custom web plugin passing its own mimetype mapping - e.g. to serve `.json`, `.ico` or `.mjs` files with a correct Content-Type (browsers refuse ES modules served as `application/octet-stream`) - gets the built-in mapping and fallback instead, with no error.

Fix: use the parameters for the lookup. No behavior change for the built-in web module, which relies on the defaults (`MIMETYPES` / `FALLBACK_MIMETYPE`).

Added a test verifying default, custom-mapping and custom-fallback behavior; fails before the fix (`text/html` returned instead of the custom type), passes after. Full test suite passes; flake8/isort/mypy clean on touched files. CHANGELOG entry added under 3.7.7.dev.

Disclosure: this fix was prepared with the assistance of an AI coding tool (Claude Code); the bug was found by code audit, and the change was human-reviewed and verified by running the test suite before and after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 3.7.7.dev
+* Fix: web plugin helpers httputils.serve_resource/serve_folder ignored their mimetypes and fallback_mimetype parameters and always used the built-in mapping, so custom web plugins could not serve additional file types with a correct Content-Type
 * Fix: free-busy REPORT always failed with HTTP 400 ("FREEBUSY occurrences limit of 0 hit") when [reporting] max_freebusy_occurrence is set to 0 (limit disabled), because the limit check did not honor the disabled limit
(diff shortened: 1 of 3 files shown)


pbiering (approved):
Thank you!

PR #2193, opened 2026-08-04
Fix: time-range filter on VTODO with DTSTART/DUE and CREATED/COMPLETED

TowyTowy:
`visit_time_ranges()` reuses a single `original_duration` variable for two unrelated purposes: the `DTSTART`→`DUE` span, and the `CREATED`→`COMPLETED` span a few lines later. When a VTODO carries all four properties — the normal shape of a *completed* task, since most clients write `CREATED` and `COMPLETED` — the second assignment clobbers the first, and the DTSTART/DUE branch of the RFC 4791 §9.9 table then reconstructs `DUE` as `DTSTART + (COMPLETED - CREATED)`.

The `elif` chain already implements the RFC table correctly (DTSTART/DUE takes precedence over CREATED/COMPLETED), so the CREATED/COMPLETED value is never wanted there. This keeps it in its own variable.

Two VTODOs, identical except that one carries `CREATED`/`COMPLETED`:

```
DTSTART:20130901T180000Z
DUE:20130903T180000Z
CREATED:20130901T170000Z     <- only in the first
COMPLETED:20130901T173000Z   <- only in the first
```

```
--- with CREATED+COMPLETED
  find_time_range -> 2013-09-01 18:00:00+00:00 2013-09-01 18:30:00+00:00
  time-range 20130902T000000Z..20130903T000000Z -> match=False
--- without CREATED+COMPLETED
  find_time_range -> 2013-09-01 18:00:00+00:00 2013-09-03 18:00:00+00:00
  time-range 20130902T000000Z..20130903T000000Z -> match=True
```

Both the `calendar-query` REPORT filter and `item.find_time_range()` (the enclosing range cached for the storage prefilter) are affected, so such a task can silently go missing from — or wrongly appear in — a client's date-range view.

Adds `radicale/tests/static/todo10.ics` and a regression test; the existing todo fixtures cover each row of the RFC table in isolation but never the combination.

Prepared with AI assistance (Claude); fail-before/pass-after verified, full suite green, flake8/isort/mypy clean on the changed files.


pbiering (approved):
Thank you!

PR #2194, opened 2026-08-07
multifilesystem: Improve performance of PROPFIND

mcrha:
The PROPFIND does not need to check the filesystem for the file existence, because the list comes from the directory listing, similar as the other parts already do it. That was highly visible with large collections (10k+ items), causing massive delays.

Also avoid repeated stat() calls on the same path.


pbiering (approved):
Good catch, thank you very much!

PR #2198, opened 2026-08-09
add systemd service

anjandev:
The project should ship a systemd service so downstream packagers can automatically adopt the latest best practices.

https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/106605/diffs#note_636736

### DIFF
--- a/contrib/systemd/radicale.service
+++ b/contrib/systemd/radicale.service
@@ -1,0 +1,28 @@
+[Unit]
+Description=Radicale CalDAV and CardDAV server
+Documentation=https://radicale.org/v3.html
+After=network.target
+Requires=network.target
+
+[Service]
+EnvironmentFile=-/etc/sysconfig/radicale
+ExecStart=/usr/libexec/radicale/radicale
+Restart=on-failure
+Type=exec
+User=radicale
+Group=radicale
+UMask=0027
+WorkingDirectory=/var/lib/radicale
+PrivateTmp=true
+ProtectSystem=strict
+ProtectHome=true
+PrivateDevices=true
+ProtectKernelTunables=true
+ProtectKernelModules=true
+ProtectControlGroups=true
+NoNewPrivileges=true
+CapabilityBoundingSet=
+ReadWritePaths=/var/lib/radicale/ /var/cache/radicale/
+
+[Install]
+WantedBy=multi-user.target


pbiering:
There are some minor differences to the one currently packaged in Fedora/EL

https://src.fedoraproject.org/rpms/radicale/blob/rawhide/f/radicale.service

Can you test this one or is there any incompatibility with Alpine Linux?

```
 [Unit]
...
+Documentation=https://radicale.org/v3.html

 [Service]
...
+EnvironmentFile=-/etc/sysconfig/radicale
+Type=exec
+Group=radicale
+WorkingDirectory=/var/lib/radicale
+CapabilityBoundingSet=
...
-ReadWritePaths=/var/lib/radicale/
+ReadWritePaths=/var/lib/radicale/ /var/cache/radicale/
```

`ReadWritePaths` should contain already both imho, will not harm but avoid issues later in case admin enables this.

anjandev:
```
commit ddf06a19cf0dc77d9a5449003aa8bb4f9c0c626a
Author: Anjandev Momi <anjan@momi.ca>
Date:   Mon Aug 10 10:38:02 2026 -0700

    use fedora radicale
    
    https://src.fedoraproject.org/rpms/radicale/blob/rawhide/f/radicale.service
    
    On Alpine, needed to make this small change because otherwise I get:
    Failed to set up mount namespacing: /var/cache/radicale: No such file or directory
    Failed at step NAMESPACE spawning /usr/bin/env: No such file or directory

diff --git a/roles/server/files/radicale.service b/roles/server/files/radicale.service
index d1715b9..2af7703 100644
--- a/roles/server/files/radicale.service
+++ b/roles/server/files/radicale.service
@@ -1,12 +1,18 @@
 [Unit]
 Description=A simple CalDAV (calendar) and CardDAV (contact) server
+Documentation=https://radicale.org/v3.html
 After=network.target
 Requires=network.target
 
 [Service]
+EnvironmentFile=-/etc/sysconfig/radicale
+Type=exec
+WorkingDirectory=/var/lib/radicale
+CapabilityBoundingSet=
 ExecStart=/usr/bin/env python3 -m radicale
 Restart=on-failure
 User=radicale
+Group=radicale
 # Deny other users access to the calendar data
 UMask=0027
 # Optional security settings
@@ -19,8 +25,6 @@ ProtectKernelModules=true
 ProtectControlGroups=true
 NoNewPrivileges=true
 ReadWritePaths=/var/lib/radicale/
-# Replace with following in case dedicated cache folder should be used
-#ReadWritePaths=/var/lib/radicale/ /var/cache/radicale/
 
 [Install]
 WantedBy=multi-user.target
```

Would you like me to push a change with the config that works on Alpine?

pbiering:
Have you tested starting Radicale directly with `/usr/libexec/radicale/radicale` instead of using `/usr/bin/env ...` ?

Fedora spec file has some tweaks and create a also a wrapper script.

https://src.fedoraproject.org/rpms/radicale/blob/rawhide/f/radicale.spec

In addition it tweaks directories later by using `sed`

I don't know whether it's a good idea to package a service file which must be tweaked by the one or the other distribution packaging mechanism to adjust paths and config. Partially it can even happen that some parameters are not supported by specific systemd version.

So either packaging a bunch of variants (Alpine, Fedora, EL) for each specific major version  into upstream or only add link to documentation where examples can be found.

And the story continues then for logrotate, sysconfig, SELinux, tmpfiles definition.

pbiering:
BTW:  In Fedora/EL the cache folder is already defined and packaged (even if unused by default config), therefore systemd service has no issue: 

```
rpm -qlv radicale3  | grep cache
drwxr-x---    2 radicale radicale                    0 Aug  6 08:45 /var/cache/radicale
```

Potentially also a suggestion for the Alpine package.

anjandev:
I have changed the pull request to follow the fedora service. It's up to the alpine maintainer if they want to make the cache folder or change the line via sed.

I do think it's a good idea for upstream to default configs despite distros' differences because it provides a baseline on how upstream expects the program to be started via systemd. You can reject nitpicks to the config. The distro maintainers can use sed or patch on the config like they would for any other file in the repo. However, providing no upstream example will lower the overall level....I copied my systemd service from the docs and I like the fedora service more.

Also, I recommend removing the example service file from the website and linking to the service file in this repo so that theres only one source of truth for maintainers and users.

https://radicale.org/v3.html#running-as-a-service

pbiering (approved):
thank you

pbiering:
Doc adjusted via c66043d27cc5edfe605b2d452ae647f2b9a31f81

PR #2201, opened 2026-08-14
Add Alpine packages to Docker image

patrickap:
This PR adds the apache2-utils and openssh-client packages to the Docker image. Both packages are required to:

1. Use a htpasswd file with encryption 
2. Use Radicale hooks to synchronize with Git over SSH

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ WORKDIR /app
 
 RUN addgroup -g 1000 radicale \
     && adduser radicale --home /var/lib/radicale --system --uid 1000 --disabled-password -G radicale \
-    && apk add --no-cache ca-certificates openssl curl git
+    && apk add --no-cache ca-certificates openssl curl git openssh-client
 
 COPY --chown=radicale:radicale --from=builder /app/venv /app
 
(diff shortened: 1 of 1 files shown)


pbiering:
@kalsi-avneet - please review

kalsi-avneet:
@patrickap , 

Thank you for the PR.

>     1. Use a htpasswd file with encryption

I believe radicale can already read a htpasswd file inside the container without installing `apache2-utils`. Could you please share your usecase that requires this installation?

patrickap:
Hi @kalsi-avneet,

Thanks for the feedback. You're right. The file can be generated and piped into the container via `docker exec` or an existing file can be mounted. However this requires the host to have the appropriate package installed. 

Including it in the image feels right to me. Ideally a container should contain everything the application needs to run. And this is a dependency required to add users via CLI. It makes the workflow more self-contained and consistent. I would say it only makes sense to have it in the container as well.

kalsi-avneet:
> Hi @kalsi-avneet,
> 
> Thanks for the feedback. You're right. The file can be generated and piped into the container via `docker exec` or an existing file can be mounted. However this requires the host to have the appropriate package installed.
> 
> Including it in the image feels right to me. Ideally a container should contain everything the application needs to run. And this is a dependency required to add users via CLI. It makes the workflow more self-contained and consistent. I would say it only makes sense to have it in the container as well.

Thank you for explaining. I understand, and agree. 

@pbiering,  this pr looks ok to me

kalsi-avneet (approved):

PR #2203, opened 2026-08-16
Minor fixes extensions related to group membership based collections

pbiering:
- improve some logging
- improve some default config examples
- add option to replace hardcoded "GROUPS" folder
- add test cases for "GROUPS" feature introduced 2024 for LDAP-only, with 3.8.0 now other authentication schemas can also take use of


pbiering:
@petervarkoly - this finally adds some test cases , as >= 3.8.0 also supports now local `htgroup` and this is a requirement to test without having LDAP server active.

Issue #2215, opened 2026-08-28
Denied access on unauthenticated public share


Whitebowfr:
When making a public R/W token share on a user's calendar, I get a 403: Forbidden when trying to access it unauthenticated. Is it a configuration issue or am I misunderstanding how sharing works ? 
The wiki specifies "token-based share (do not require user authentication)", but reading the entire documentation I can't find anything else talking about non-authenticated users.

Here are the relevant logs : 
```shell

radicale-1  | [2026-08-28 10:31:26 +0000] [1/Thread-9 (process_request_thread)] [INFO] PUT request for '/.token/v1/bJkMMqFpSjGVwwUqy0AcDgAmnHmL_3FPM686CoBlThnm/b4c42e97-6a4c-47a2-ab59-1a4171149ff6.ics' received from 10.10.6.3 (forwarded for '90.119.206.209') using 'Mozilla/5.0 (X11; Linux x86_64; rv:153.0) Gecko/20100101 Thunderbird/153.0'
radicale-1  | [2026-08-28 10:31:26 +0000] [1/Thread-9 (process_request_thread)] [NOTICE] Access to '/.token/v1/bJkMMqFpSjGVwwUqy0AcDgAmnHmL_3FPM686CoBlThnm/b4c42e97-6a4c-47a2-ab59-1a4171149ff6.ics' denied
radicale-1  | [2026-08-28 10:31:27 +0000] [1/Thread-9 (process_request_thread)] [INFO] PUT response status for '/.token/v1/bJkMMqFpSjGVwwUqy0AcDgAmnHmL_3FPM686CoBlThnm/b4c42e97-6a4c-47a2-ab59-1a4171149ff6.ics' in 1.263 seconds gzip 61 bytes: 403 Forbidden

```

Here is my config file : 
```conf
[auth]
type = htpasswd
htpasswd_filename = /etc/radicale/users
htpasswd_encryption = autodetect

[sharing]
type = files
permit_create_token = True
permit_create_map = True
permit_properties_overlay = True
collection_by_token = True
collection_by_map = True

[rights]
type = from_file

```

And here is my rights file : 
```conf
# Allow reading root collection for authenticated users
[root]
user: .+
collection:
permissions: R

# Allow reading and writing principal collection (same as username)
[principal]
user: .+
collection: {user}
permissions: RW

# Allow reading and writing calendars and address books that are direct
# children of the principal collection
[calendars]
user: .+
collection: {user}/[^/]+
permissions: rw

```

[...]

PR #2218, opened 2026-08-31
Fix public token share issues

Whitebowfr:
Fixes the issue #2215 by removing the end-of-line selector to the regex check for supported tokens (allowing the calendar path to pass through), and by adding back the calendar path after the token resolver (prevents an issue that popped up after the first fix, where the suffix is missing, and thus it tries to overwrite the calendar and throws a HTTP_IF_NOT_MATCHED error)

I added a test case to PUT an event in a token-shared calendar.


pbiering:
@Whitebowfr - thank you for your PR, unfortunatly, there are more so far undetected corner cases...will replace your PR by a new one.

PR #2228, opened 2026-09-13
Add optional caldav-server-tester based compatibility test suite.

### LINKED ISSUE #1911: Compatibility checking in test suite
I was promising this before the summer, but suddenly other things started happening in my life preventing me from focusing on the caldav project.  Anyway, now I've released the caldav client library version 2.1 and version 0.1 of my caldav-server-tester project - so it should now be possible to utilize it in the Radicale test suite.  I think it's nice as it may prevent compatibility regression issues.  It may also uncover other compatibility issues in future releases of the caldav-server-tester ... which may be nice or an annoyance, dependent on your point of view :-)

To call the caldav-server-tester from the Radicale test suite, it's needed to:

* add the caldav-server-tester and caldav to the optional dependencies of the project
* construct a `client` instance of the `caldav.davclient.DAVClient` class that can talk with the calendar system. This either means:
  * Starting up a Radicale server on localhost
  * Mock up the `client.request`-method so it will call on the appropriate radicale methods without having to go via IP.
 
The actual test code would look something like this (see also `testCheckCompatibility` in `caldav/tests/test_caldav.py`):

```python
from caldav_server_tester import ServerQuirkChecker
checker = ServerQuirkChecker(self.caldav, debug_mode='assert')
checker.check_all()
observed = checker.features_checked.dotted_feature_set_list(compact=True)
expected = ...
assert(observed == expected)
```

The testCheckCompatibility in the caldav project also has quite some logic to smoothen out minor deviations between the observed and the expected, not sure if the Radicale project would need something like that.

If you want me to, I may try to write a pull request for this test, but possibly I may need some help with it.

Currently, the expected compatibility issues with Radicale looks like this:

```
{
    "search.category.fullstring": {"support": "unsupported"},
    "search.recurrences.includes-implicit.todo.pending": {"support": "unsupported"},
    "search.recurrences.expanded.todo": {"support": "unsupported"},
    "search.recurrences.expanded.exception": {"support": "unsupported"},
 }
```

I do have an old list of other known issues with Radicale, but they are not covered by the `caldav-server-tester` yet.

The `search.category.fullstring`-feature holds true if an event with `CATEGORIES: foo,bar,zoo` will be returned when doing a search for `foo,bar,zoo`. 

`search.recurrences.includes-implicit.todo.pending` is a weird bug that I've found only with Radicale.  Give a recurring todo (say, yearly event starting in year 2024) with `STATUS:NEEDS-ACTION`, and then make a query asking for tasks for 2025, and the task will be returned as it should.  Make a query asking for tasks for 2024 with status `NEEDS-ACTION` and the task will be returned.  So far all good - but then make a query for tasks for 2025 with status `NEEDS-ACTION`, and the event will not be returned.  I will try to get time to create a separate issue for this, complete with the icalendar data and the XML used in the report-query.

Apparently server-side expanding of events work, while server-side expanding of tasks does not work.  And server-side expanding of events with exceptions (a `VEVENT` with the same `UID` and a `RECURRENCE-ID`) also doesn't seem to work.  Please tell if you want me to file separate issues for those.

Are those compatibility issues listed above important?  No, most likely not:

* search.category.fullstring is unimportant.  There is not much point in searching for events with `foo,bar,zoo` - an event with `CATEGORIES:bar,zoo,foo` should be considered to be equivalent as far as I understand the icalendar RFC.
* I have a feeling that my plann project is the only real world user of "complicated server-side caldav searches".  While the bug above does affect the plann project, there exists a workaround now in version 2.1 of the caldav library.
* If I remember correct, the RFC says that expanding SHOULD work, but many caldav servers doesn't support expanding at all, it's probably another caldav feature that isn't much used in the wild.  My caldav library is by now by default doing the expand work client-side. so I'm not affected by it.

theofficialtruck:
## Summary

Fixes #1911. Integrates the `caldav` client library and the
`caldav-server-tester` project into Radicale's test suite for CalDAV
compatibility checking, as proposed by @tobixen (maintainer of both
`caldav` and `caldav-server-tester`) in the issue.

- New optional `caldav_test` extra (`caldav>=3.0`,
  `caldav-server-tester>=1.2,<2`) — opt-in, mirroring how the Playwright
  integration tests are isolated to their own `dev` extra, so it doesn't
  become a hard dependency of the default `pytest` run.
- New `caldav_compat_tests/` directory:
  - `common.py` starts a real Radicale server as a subprocess on a free
    localhost port. A real HTTP server is required here — `DAVClient`
    makes genuine network requests, so the existing in-process WSGI test
    client (`radicale/tests/__init__.py`) can't be reused for this.
  - `test_compatibility.py` points `caldav_server_tester.ServerQuirkChecker`
    at that server, runs `check_all()`, and asserts the observed
    feature-support deviations match a documented `EXPECTED_DEVIATIONS`
    baseline. A mismatch means either a compatibility regression or an
    intentional change that requires updating the baseline.
- New `caldav_compat_test` tox env and `caldav-compat-test` CI job
  (Ubuntu only, same as the existing `integ-test` job).
- `README.md` explaining how to run the suite and update the baseline.

## Notes on the expected-deviations baseline

The issue is from late 2023/2025 and quotes an expected-deviations dict
from `caldav-server-tester` 0.1. I re-ran the checker (v1.2.0, the current
release) against a live build of this branch rather than trust the stale
dict — the check names and coverage have changed substantially since then,
and several of the originally-reported gaps are already fixed:

- `search.recurrences.includes-implicit.todo.pending` — now full support
- `search.recurrences.expanded.exception` — now full support
- `search.category.fullstring` — the check itself no longer exists in
  1.2.0; the closest current equivalents (`search.text.category`,
  `search.is-not-defined.category`) report full support

The current baseline instead reflects:
- Radicale doesn't implement CalDAV Scheduling (RFC 6638) or
  principal-search (`scheduling*`, `principal-search*`,
  `get-current-user-principal.has-calendar`)
- `search.recurrences.expanded.todo` — expanding recurring tasks
  server-side is still unsupported
- `search.text.case-sensitive` — case-sensitive (`i;octet`) text search
  collation isn't distinguished from the default

None of these looked like something this PR should also try to fix —
happy to split any of them into follow-up issues if maintainers want.

## Test plan

- [x] `pytest caldav_compat_tests -v` passes locally against a live
      Radicale build
- [x] `flake8`, `isort --check`, `mypy` all clean on the new files
- [x] Confirmed `pytest .` from repo root still collects 442 tests
      (unchanged) — the new directory is properly excluded via
      `norecursedirs`
- [x] CI run on this PR (new `caldav-compat-test` job)


pyproject.toml
```diff
@@ -220,21 +221,28 @@ setenv = { PLAYWRIGHT_BROWSERS_PATH = "{toxworkdir}/.playwright" }
 commands_pre = [["playwright", "install", "chromium"]]
 commands = [["pytest", "integ_tests"]]
 
+[tool.tox.env.caldav_compat_test]
+extras = ["caldav_test"]
+deps = [
```
pbiering, line 226: I was only able to run it locally by applying following extension:

```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -224,6 +224,8 @@ commands = [["pytest", "integ_tests"]]
 [tool.tox.env.caldav_compat_test]
 extras = ["caldav_test"]
 deps = [
+    "caldav",
+    "caldav_server_tester",
     "pytest"
 ]
 commands = [["pytest", "caldav_compat_tests"]]
```

Please incorporate, thank you!

pbiering (changes requested):

tobixen:
v1.3.0 will be launced in some few days.

theofficialtruck:
@pbiering i have added caldav and caldav-server-tester explicitly to deps in the caldav_compat_test tox env, matching how integ_test lists pytest-playwright directly. i have also verified this locally using both plain pytest and the actual tox -e caldav_compat_test command on Linux. please re-review, thanks!

theofficialtruck:
@pbiering Thanks for the review! Since then, I have pushed a few more fixes, including your original ask; all of which are specified below:

1. Added `caldav` and `caldav-server-tester` explicitly to `deps` in the `caldav_compat_test` tox env, matching how `integ_test` lists `pytest-playwright` directly (this was your original ask).
2. Found and fixed a `uv run` dependency resolution failure that was also breaking `integ-test`, not just the new job: `caldav>=3.0` requires Python>=3.10, but this project supports `>=3.9`, and `uv` resolves one dependency graph valid across the whole declared Python range regardless of which extra is requested. Marked the `caldav/caldav-server-tester` deps with `python_version >= 3.10` and added `pytest` to the extra. This was missing, so `uv run --extra caldav_test pytest ...` had no guaranteed way to find it.
3. Fixed a pre existing, unrelated flake on the PyPy matrix job: several `os.scandir()` calls elsewhere in the codebase (`sharing/files.py`, `storage/multifilesystem/{cache,discover,get,history}.py`) were not using a context manager, unlike the fix for #1972 in `pathutils.py`. PyPy's non-refcounting GC surfaced the leftover leak as a `ResourceWarning` attributed to an unrelated test.

Please re review, thanks!

pbiering (approved):
thank you

PR #2230, opened 2026-09-20
Add KashCal to the list of supported Android clients

one-kash:
KashCal is an open-source (Apache-2.0) Android calendar app with native CalDAV/CardDAV sync. It is tested against Radicale and works with it, so I've added it alongside DAVx⁵ in the Supported Clients list.

- Source: https://github.com/KashCal/KashCal
- F-Droid: https://f-droid.org/packages/org.onekash.kashcal


pbiering (approved):
thank you

PR #2231, opened 2026-09-21
Add download button for mobileconfig for Apple devices

maxberger:
* Created unified helper for downloads with password
* Added Icon button to download mobileconfig
* Added integration tests


maxberger:
caldav compat test has been updated very recently: https://github.com/python-caldav/caldav-server-tester

Since I did not change any server code, I am assuming the issues are unrelated.

pbiering:
@theofficialtruck - can the caldav version somehow be pinned by tagged for a planned change and not using latest?

https://github.com/python-caldav/caldav-server-tester/tags

theofficialtruck:
> @theofficialtruck - can the caldav version somehow be pinned by tagged for a planned change and not using latest?
> 
> 
> 
> https://github.com/python-caldav/caldav-server-tester/tags

Yes this is possible. I reproduced the failure on current master: the compat test passes with caldav-server-tester 1.2.0 and fails with 1.3.0, so its unrelated to this PR. We can change the extra to an exact pin (1.2.0) and bump it deliberately, updating EXPECTED_DEVIATIONS when we do. Happy to open a PR for this.

pbiering (approved):
Thank you!

pbiering:
> > @theofficialtruck - can the caldav version somehow be pinned by tagged for a planned change and not using latest?
> > https://github.com/python-caldav/caldav-server-tester/tags
> 
> Yes this is possible. I reproduced the failure on current master: the compat test passes with caldav-server-tester 1.2.0 and fails with 1.3.0, so its unrelated to this PR. We can change the extra to an exact pin (1.2.0) and bump it deliberately, updating EXPECTED_DEVIATIONS when we do. Happy to open a PR for this.

Please create a PR, best would be having the specific version specified in `.github/workflows/test.yml` as parameter

